PR #84, opened 2014-10-03
Updated OpenVPN Configuration Guide for CentOS 7

Dakta:
@movesmyers @displague re: [Secure Communications with OpenVPN on CentOS 6](https://www.linode.com/docs/networking/vpn/secure-communications-with-openvpn-on-centos-6)

After spending a day getting OpenVPN running on CentOS 7, I've provided an upgraded version of the CentOS 6 guide for users who wish to do the same. This should not be considered new work, merely an updated version of an existing guide. However, if you'd like to compensate me for putting it together, a small credit to my Linode account (same name) would be appreciated.

Here is a summary of my major changes.

Best practice for enabling EPEL is to install it through yum itself, not directly with an RPM:

    yum instal epel-release

As of CentOS 7, there have been changes to the networking subsystem and other core functionality which should be noted.

First, iptables has been replaced by [firewalld](https://fedoraproject.org/wiki/FirewallD), so old-style iptables configuration is not supported "out of the box", rendering the suggested iptables configuration impossible. This can be remedied by installing iptables-services.

Also, https://www.linode.com/docs/networking/vpn/secure-communications-with-openvpn-on-centos-6#connect-to-the-openvpn is no longer accurate, and would lead even a reasonably experienced user to frustration as openvpn no longer uses init.d scripts. Any attempt to find documentation for this problem online will lead to frustration, and any attempt to use a previous version of an init.d script for openvpn will also fail due to changes in the networking subsystem. CentOS 7 now uses systemctl, which requires a slightly different startup command, as documented in the ArchLinux wiki: https://wiki.archlinux.org/index.php/OpenVPN#systemd_service_configuration

TODOs:

- Update iptables configuration to reflect the recommended configuration described in https://www.linode.com/docs/security/securing-your-server. Currently there will be conflicts with ports and protocols being open, as well as NAT routing VPN traffic out of the system. This isn't terrible if you're on a new install and don't mind flushing your iptables rules and trying again a few times with rules from other guides from the internet, but it could be problematic for users on established systems.

- Update iptables configuration to use now-current, recommended firewalld.


alexfornuto:
I'm going to close this PR has it has not received any updates. Please feel free to create a new PR when you've resolved the issues discussed.

Issue #105, opened 2014-10-26
Dovecot mail_location and virtual user home directory should not be set to the same value


deoren:
Hi,

First of all, I want to say thanks to the author of the [Email with Postfix, Dovecot, and MySQL](https://www.linode.com/docs/email/postfix/email-with-postfix-dovecot-and-mysql) guide. It was easy to follow and got me started on the road from Dovecot + local system users setup to one where I can easily manage users through a database. I've since been able to customize and tweak the setup to fit all of my needs. 

That said, there is an issue with the guide that didn't bite me until I upgraded to Dovecot 2.2.x about a week ago. I've since learned that it is a bad idea to have the `mail_location` set to the same location as the virtual user's `home` directory:

Quoting the official [Home Directories for Virtual Users](http://wiki2.dovecot.org/VirtualUsers/Home) Dovecot wiki page:

>Home vs. mail directory
>
> Home directory shouldn't be the same as mail directory. It's possible to do that, but you might run into trouble with it sooner or later.

So with the guide directing the reader to configure their setup like so:

```
userdb {
   driver = static
   args = uid=vmail gid=vmail home=/var/mail/vhosts/%d/%n
}
```

`mail_location = maildir:/var/mail/vhosts/%d/%n`

The user will have their Maildir content dumped right into the root of their Dovecot home directory. 

Some additional back & forth on the Dovecot mailing list (in case it's useful):

* http://dovecot.org/pipermail/dovecot/2014-October/098379.html
* http://dovecot.org/pipermail/dovecot/2014-October/098400.html

In my case when I used `doveadm expunge` from a cron job prior to the upgrade I didn't have any problems. It was only after the upgrade from v2.0.19 (default in Ubuntu 12.04 LTS) to v2.2.x that doveadm threw errors about finding `.dovecot.lda-dupes` files in the home directory for some of our virtual user's.

It didn't turn out to be anything serious from what I can tell, but it was enough to highlight that worse could happen at some point due to the configuration I was using. I propose that the guide be modified to recommend this configuration instead:

```
userdb {
   driver = static
   args = uid=vmail gid=vmail home=/var/mail/vhosts/%d/%n
}
```

`mail_location = maildir:~/Maildir`

The `userdb` setting would stay the same, but the setting for `mail_location` would be changed from a static value to one that inherits the virtual user's home directory *and* places mail into a subdirectory. This would match the recommended values in the Dovecot configuration and should prove a safer set of default values for readers to adopt.

[...]

PR #114, opened 2014-11-08
Update Dovecot mail_location value to store mail within a subdirectory inside of the user's mail home directory

deoren:
As discussed on issue #105, here are the changes to one of the guides that sets the `mail_location` value to a subdirectory within the mail home directory. This prevents some of the issues mentioned on the official [Home Directories for Virtual Users](http://wiki2.dovecot.org/VirtualUsers/Home) Dovecot wiki page and also discussed on the mailinglist thread highlighted by these two posts:

* http://dovecot.org/pipermail/dovecot/2014-October/098379.html
* http://dovecot.org/pipermail/dovecot/2014-October/098400.html

There are additional changes to the [https://www.linode.com/docs/email/postfix/troubleshooting-problems-with-postfix-dovecot-and-mysql#virtual-domains-and-users](Troubleshooting Problems with Postfix, Dovecot, and MySQL) that are being addressed on PR #113 and a future PR.

P.S.

I didn't know what the official recommendation/practice was regarding merging in upstream repo changes prior to making a PR, so I made the assumption that I would need to do so in order for the changes to be accepted. Perhaps I'm just thinking of the CVS/SVN culture and showing fewer commits to a bugfix branch is the better way to go.


buhman:
> the mail will be stored in the Maildir subdirectory within the virtual users's home

SGTM.

Postfix is using dovecot's lmtp, so no other changes are needed to make the article consistent.

:thumbsup: should merge.

alexfornuto:
I agree with @buhman, and there are concerns with giving a system user access to home dirs.

PR #135, opened 2014-12-16
Updated distro-supplied kernel guide for Fedora 21

Pacoup:


docs/tools-reference/custom-kernels-distros/run-a-distributionsupplied-kernel-with-pvgrub.md
```diff
@@ -564,46 +564,58 @@ CentOS 5
...
-7.  Reboot your Linode from the **Dashboard** tab.
-8.  Once your Linode has rebooted, log in via SSH and issue the following command:
+5. Create a file named `/boot/grub/menu.lst` with the following contents, replacing values to match the contents of the `/boot` directory:
 
-        uname -a
+        timeout 5
+        title Fedora 21 (3.17.6-300.fc21.x86_64)
+        root (hd0)
+        kernel /boot/vmlinuz-3.17.6-300.fc21.x86_64 root=/dev/xvda
+        initrd /boot/initramfs-3.17.6-300.fc21.x86_64.img
```
buhman: should probably keep using `{: .file }`

docs/tools-reference/custom-kernels-distros/run-a-distributionsupplied-kernel-with-pvgrub.md
```diff
@@ -564,46 +564,58 @@ CentOS 5
...
-1.  Make sure your package repositories and installed packages are up to date by issuing the following command:
+1. Make sure your package repositories and installed packages are up to date by issuing the following command:
 
         yum update
 
-2.  Issue the following command to install the default kernel for Fedora 13:
+2. Issue the following command to install the default kernel for Fedora 21:
 
-    **32-bit Fedora:** :
+        yum install kernel.x86_64
```
buhman: I suggest dropping the `.x86_64` to make this architecture-agnostic.

docs/tools-reference/custom-kernels-distros/run-a-distributionsupplied-kernel-with-pvgrub.md
```diff
@@ -564,46 +564,58 @@ CentOS 5
...
+8. Once your Linode has rebooted, you can confirm that your new kernel is running by logging in to ssh and issuing the following command:
 
-Note that if you later install an updated kernel, you'll need to add an entry for it to your `menu.lst` file. By default, the first kernel in the list will be booted.
+        uname -a
 
+   You should see output similar to the following:
 
+        Linux dz8 3.17.6-300.fc21.x86_64 #1 SMP Mon Dec 8 22:29:32 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux
 
+Note that if you later install an updated kernel, you'll need to add an entry for it to your `/boot/grub/menu.lst` file.
```
buhman: Most unfortunate.

Better would be to chain grub-xen, and avoid writing menu.lst entirely.

docs/tools-reference/custom-kernels-distros/run-a-distributionsupplied-kernel-with-pvgrub.md
```diff
@@ -564,46 +564,58 @@ CentOS 5
 
         Linux li181-194 2.6.18-194.26.1.el5xen #1 SMP Tue Nov 9 14:13:46 EST 2010 i686 i686 i386 GNU/Linux
 
-Fedora 17
+Fedora 21
 ---------
 
-1.  Make sure your package repositories and installed packages are up to date by issuing the following command:
+1. Make sure your package repositories and installed packages are up to date by issuing the following command:
```
alexfornuto: The extra space after the number is required for our site to properly see the command below without any extra spaces prepended to it.

docs/tools-reference/custom-kernels-distros/run-a-distributionsupplied-kernel-with-pvgrub.md
```diff
@@ -564,46 +564,58 @@ CentOS 5
...
 
         yum update
 
-2.  Issue the following command to install the default kernel for Fedora 13:
+2. Issue the following command to install the default kernel for Fedora 21:
```
alexfornuto: Same issue.

docs/tools-reference/custom-kernels-distros/run-a-distributionsupplied-kernel-with-pvgrub.md
alexfornuto: Agreed. Please see https://www.linode.com/docs/tools-reference/tools/markdown-quick-reference#file-excerpt

Pacoup:
Alright, thanks for the comments. I will fix the layout issues soon and try my hand at setting up grub-xen chaining.

Pacoup:
@buhman Well, I have not figured out how to do the whole grub-xen chaining. I don't know enough about Xen and the Linux boot process to write something adequate for this. It'd be great if you could help though, because having to go and update the kernel version on every kernel update is not so great.

@alexfornuto Hopefully the spaces and ```{: .file }``` stuff are okay now. Also, since this is basically the same guide with the same caveats as the previous one for Fedora 17, why not pull this and update the guide with some better solution (grub-xen chaining?) in a separate PR?

buhman:
@Pacoup, you'll need to build grub-xen, which looks something like this:

```sh
yum install make automake gcc bison flex xz-devel xen-devel
git clone git://git.sv.gnu.org/grub.git
cd grub
./autogen.sh
./configure --prefix=/usr --target=x86_64 --with-platform=xen
make V=1 -j$(nproc)
```

Doesn't take long, even on a Linode 1GB.

Next build a 'standalone' grub-xen image with an embedded configuration that loads Fedora's grub config.

```sh
cat <<EOF > boot/grub/grub.cfg
search -s root -f /boot/grub2/grub.cfg
configfile /boot/grub2/grub.cfg
EOF
./grub-mkstandalone --grub-mkimage=./grub-mkimage -o grub.xen -O x86_64-xen -d grub-core/ boot/grub/grub.cfg
```

To keep things simple(er?), the upstream `grub2-tools` can be used for subsequent grub.cfg building. However Fedora's ``10_linux`` isn't quite right, so we should fix that first:

```sh
yum install kernel grub2-tools
sed -i '/sixteenbit="16"/d' /etc/grub.d/10_linux
grub2-mkconfig -o /boot/grub2/grub.cfg
```

Now we just need to tell pv-grub about grub-xen:

```sh
mkdir /boot/grub
cat <<EOF> /boot/grub/menu.lst
title grub
root (hd0)
kernel /boot/grub/grub.xen
EOF
cp grub.xen /boot/grub
```

Do the thing with the config profile, and everything ShouldWork™.

Let me know what you think.

AtOMiCNebula:
(note: I'm Nebula from https://forum.linode.com/viewtopic.php?f=20&t=10128)

Grubby, which is often installed by default on Fedora, automatically updates grub boot configurations for you whenever you install new kernels.  I had followed these instructions back in the day, and aside from some initial setup fun, have never needed to manually modify any of grub's files.  Food for thought. :)

Anyways, since I'm way behind the times and was still running a 32-bit Fedora 20, I'm just now getting around to joining the 64-bit world with a fresh Fedora 21 Linode deployment.  I'll follow these steps and report back on whether grubby continues to do the trick (or not).

AtOMiCNebula:
Sleuthed this one out.  Turns out there are a few issues going on here (and I answered the "I don't know why grubby doesn't work the first time" question, too). :)

Steps I took to get a fresh F21 deployment using a F21 kernel:

1. `mkdir /boot/grub`
2. Save contents of below snippet to `/boot/grub/menu.lst`
3. `ln -s ../boot/grub/menu.lst /etc/grub.conf`
4. `touch /boot/initial` (to make Grubby happy, see below)
5. `ln -s /dev/xdva /dev/root` (also to make Grubby happy; replace xvda with your real root device, as needed)
6. `yum install kernel`
7. Remove now-unnecessary "Initial" configuration from menu.lst, update default number, then `rm /boot/initial`

A little commentary on "making Grubby happy": as I mentioned before, Grubby is a Fedora tool to keep boot configs up to snuff whenever `yum` installs a new kernel.  However, out of an abundance of caution, it lives a very paranoid life, due to it not wanting to wreck your boot configs.  This caution leads to two issues that need to be resolved:

- Grubby will not consider a boot entry valid if it cannot see the kernel file that is being referred to, hence the `touch /boot/initial` to get us started.
- Grubby will also not consider a boot entry valid if filesystem UUIDs don't line up.  This one is dumb, but [to make a long story short](https://bugzilla.redhat.com/show_bug.cgi?id=833011), `/proc/mounts` refers to `/dev/root` when the kernel is booted without an initrd/initramfs, but the default Fedora environment doesn't put anything at `/dev/root`.  Symlinking `/dev/root` to your real root device (i.e. `/dev/xvda`) helps Grubby maintain a normal blood-pressure level.

Once Grubby is happy, installing a new kernel triggers the right menu.lst updates, and all that's left is to clean out your initial crap you put in there. :)

Happy to submit updated commits for this one.

##### Boilerplate menu.lst
Adjust to your own liking, but this base config (specifically, the kernel args) have been working fine for me for years.
```
default=0
timeout=5

title Initial
        root (hd0)
        kernel /boot/initial root=/dev/xvda console=tty0 console=hvc0 ro quiet LANG=en_US.UTF-8
```

alexfornuto:
@Pacoup you're current standing changes look good. If you're planning on looking into the chaining thing I'll leave this open, otherwise we can merge this, and you could always make a new PR down the road. What is your preference?

docs/tools-reference/custom-kernels-distros/run-a-distributionsupplied-kernel-with-pvgrub.md
```diff
@@ -564,46 +564,60 @@ CentOS 5
...
 
         uname -a
 
-    You should see output similar to the following, indicating you're running the native kernel:
-
-        Linux li63-119 3.9.10-100.fc17.x86_64 #1 SMP Sun Jul 14 01:31:27 UTC 2013 x86_64 x86_64 x86_64 GNU/Linux
-
-Note that if you later install an updated kernel, you'll need to add an entry for it to your `menu.lst` file. By default, the first kernel in the list will be booted.
-
+   You should see output similar to the following:
```
alexfornuto: Not enough indent here to line everything up under the final step.

Pacoup:
@alexfornuto I think I still want to look at chaining, cause it's so much better that way. I've been busy with other stuff, but I still intend to have a look at this soon.

alexfornuto:
OK, we'll leave this PR alone until we hear back from you.

Pacoup:
So this new version uses @AtOMiCNebula's grubby setup which I quite liked. Works and tested on both Fedora 20 and Fedora 21. The setup also works when upgrading from Fedora 20 to Fedora 21, so it looks like grubby is properly maintaining the kernel version in menu.lst.

AtOMiCNebula:
Wooo! :)

AtOMiCNebula:
One additional step is needed for `grubby` to work best, at least on my VM: the following needs to be added to `/etc/sysconfig/kernel`, if it's not already there:

    UPDATEDEFAULT=yes
    DEFAULTKERNEL=kernel

Otherwise, `grubby` will not mark newly installed kernels as default.  I don't know what normally writes this file this way...maybe `anaconda`?

Pacoup:
Interesting. It's there by default on Linode's current Fedora 20 image, but not on Fedora 21.

```BASH
# UPDATEDEFAULT specifies if new-kernel-pkg should make
# new kernels the default
UPDATEDEFAULT=yes

# DEFAULTKERNEL specifies the default kernel package type
DEFAULTKERNEL=kernel
```

What would be a good way to test a kernel upgrade on Fedora 21 if everything is already up to date?

AtOMiCNebula:
I validated my above suggestion worked by uninstalling `kernel`/`kernel-core`/`kernel-modules`, making the changes, and then reinstalling `kernel` (which of course pulls in `kernel-core`/`kernel-modules` as dependencies).  When you uninstall the kernel, `grubby` removes the options from the list, and after reinstalling, you should see the default pointing to the new kernel entry.

Pacoup:
(Two decades later) Right, this should be better. Thanks @AtOMiCNebula

thetechstewart:
Due to this guide being superseded by our KVM distro kernel guide (https://www.linode.com/docs/tools-reference/custom-kernels-distros/run-a-distribution-supplied-kernel-with-kvm) I'm marking this PR closed.

Issue #136, opened 2014-12-17
email should be unique per domain in "virtual_users" table [Email with Postfix, Dovecot, and MySQL]


mtantawy:
I was reading the tutorial **"Email with Postfix, Dovecot, and MySQL"** https://www.linode.com/docs/email/email-with-postfix-dovecot-and-mysql#creating-the-database and i think something in the step of creating the database tables can be done in a better way.

in **"Creating the Database"** https://www.linode.com/docs/email/email-with-postfix-dovecot-and-mysql#creating-the-database
The table *"virtual_users"* is created as such
```sql
CREATE TABLE `virtual_users` (
  `id` int(11) NOT NULL auto_increment,
  `domain_id` int(11) NOT NULL,
  `password` varchar(106) NOT NULL,
  `email` varchar(100) NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `email` (`email`),
  FOREIGN KEY (domain_id) REFERENCES virtual_domains(id) ON DELETE CASCADE
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

I noticed that the field *"email"* is marked as unique, i think this is not the best thing to do, because this forces email *"xyz"* to be unique across all domains, which should not be the case.

So, i propose the unique key to be a combination of *"email"*, and *"domain_id"* fields, this way the email uniqueness is enforced only over the domain, so an email of *"xyz"* can exist in more than one domain naturally.

So what do you think? and what are the steps should i take to apply this change? and should i apply it to all versions of this tutorial or only the main one?

PR #138, opened 2014-12-20
email should be unique per domain in "virtual_users" table

mtantawy:
Changing unique key in table *'virtual_users'* to be combination of *'domain_id'* and *'email'* to allow for email duplication across different domains **but** not in the same domain.

more details can be found in the issue i opened here, https://github.com/linode/docs/issues/136

### DIFF
--- a/docs/email/postfix/email-with-postfix-dovecot-and-mysql.md
+++ b/docs/email/postfix/email-with-postfix-dovecot-and-mysql.md
@@ -208,7 +208,7 @@ Here's how to create the necessary database and tables in MySQL:
           `password` varchar(106) NOT NULL,
           `email` varchar(100) NOT NULL,
           PRIMARY KEY (`id`),
-          UNIQUE KEY `email` (`email`),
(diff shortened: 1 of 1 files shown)


buhman:
There are several inconsistencies in this article, so I understand confusion.

First have a look at the [password_query](https://github.com/linode/docs/pull/138/files#diff-ad14e60055811ab7a2093ca374d27b68R886) used in this article. [PasswordDatabase](http://wiki2.dovecot.org/PasswordDatabase) tells us that we should only return a `user` column if we want to tell Dovecot the real `user` is something other than what we queried for, and that `user` is a combination of `username` + `domain`.

[Variables](http://wiki2.dovecot.org/Variables) tells us that the `%u` variable in the query string that we're using is indeed also `username@domain`.

`%s` in the postfix query string works the [same way](http://www.postfix.org/mysql_table.5.html).

:thumbsdown: should not merge.

mtantawy:
Seems like i misunderstood that bit, sorry for any inconveniences, and thanks for the explanation :)

PR #139, opened 2014-12-22
Apache SSO  with LDAP

alexfornuto:

### DIFF
--- a/docs/websites/apache-tips-and-tricks/apache-sso-ldap.md
+++ b/docs/websites/apache-tips-and-tricks/apache-sso-ldap.md
@@ -1,0 +1,260 @@
+
+1. Introduction
+
+This article gives an overview of the process involved in configuring Apache 2.4 for SSO with mod_auth_form, mod_session and LDAP. The SSO is completly managed on the Apache side without any involvement from the backend application(s).
+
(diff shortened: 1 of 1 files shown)


buhman:
:thumbsdown:

The entire idea also seems extremely specialized--I suspect article maintenance isn't worth the very small number of people this might interest. Also needs several adjustments to tone/point of view/style prior to publishing.

*Assumes an ldapd already exists, which is quite unlikely*. A series on ldap might be a good idea, and luckily upstream has [already](http://www.openldap.org/doc/admin24/quickstart.html) [done](http://www.openldap.org/doc/) all of the work for that.

This also uses mod_session--unless people have similarly specialized applications, this won't be very directly useful to people wanting to use said session in the application. If the session generated by mod_session isn't used by the application, most of the work done in the guide is fairly useless.

A more generic guide on "restricting access via (session-less) http basic authentication" is likely a better idea, applicable to a far greater number of people, has fewer prerequisites, and is easier to explain/do.

alexfornuto:
Thanks for the input @buhman. If there are no other  thoughts on this soon I'll close it out.

ryanwohara:
We'd be better off with LDAP guides than this.

alexfornuto:
Thanks for your input @orzFly @buhman and @ryanwohara. In light of your comments I'm going to close this PR.

PR #142, opened 2014-12-27
Make changes to IPv6 pool to add forwarding information

drussell393:
We shouldn't be using ifconfig (deprecated). We also should be removing ip addr add from this, because the changes won't be saved permanently unless they're in the /etc/network/interfaces file (or similar, depending on distribution).

I've kept the second IPv6 address add section, but removed the first because it's unnecessary to have the same information twice in the same documentation. 

While ArchLinux uses systemd-networkd now, there's a lot of people still using netctl, so I believe that we'll be okay keeping netctl in there, but will likely add systemd-networkd as well shortly. This change is more of a "pushing it out there because adding IPv6 by ip addr add is a bad idea" kind of change. 

Additionally, I've added information about IPv6 forwarding. While I don't want to add information about IPv6 forwarding at a kernel/sysctl level (eg. making your Linode a router, essentially), I do want to add a bit about /116 pools.

### DIFF
--- a/docs/applications/development/developing-git-github.md
+++ b/docs/applications/development/developing-git-github.md
@@ -15,11 +15,7 @@ published: 'Friday, February 6th, 2015'
 title: Developing with Git and Github
 ---
 
-<<<<<<< HEAD
-Git is a version control system, or VCS, a VCS tracks file revisions. Github is a website that allows collaboration between developers who use Git. With Git and Github, programmers from across the world can share ideas and code in an organized and up-to-date process.
-=======
 Git is a file tracking application, or version control system. Github is a website that allows collaboration between developers who use Git. With Git and Github, programmers from across the world can share ideas and code in an organized and up-to-date process.
->>>>>>> f22b994562f17dd40aa37964766ed31a170b2e36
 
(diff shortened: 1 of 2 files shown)


drussell393:
Working on this to make necessary adjustments after second look at it.

drussell393:
This is nearly done. I'm waiting for @buhman to provide assistance with two different things at this point.

1) Does NetworkManager properly handle Router Advertisements? If so, how2IPv6Pool on CentOS 7/Fedora 20 properly.

If not, do we advise switching back to network.service for IPv6 pools?

2) How2IPv6Pool on non-netctl Arch Linux/Fedora 21. (Arch Linux has little to no docs about systemd-networkd IPv6 static config)

I'll speak with him as soon as I can, then touch up this guide and we'll be g2g.

drussell393:
@alexfornuto This is good to go and waiting for docs review :smile:

drussell393:
@alexfornuto This should be g2g

drussell393:
Fixed the conflicts.

alexfornuto:
Awesome, Thanks @drussell393 !

PR #143, opened 2014-12-27
Removed aliases for Debian and CentOS as per iproute2

drussell393:
We should not be using aliases anymore. After ifconfig was deprecated and iproute2 took precedence, aliases became a thing of the past. 

As per Debian: "The newer ip utility does not use the same concept of aliases or virtual interfaces." 

Using aliases doesn't break the configuration, but it's probably not the best practice. It's best not to use them.

### DIFF
--- a/docs/networking/linux-static-ip-configuration.md
+++ b/docs/networking/linux-static-ip-configuration.md
@@ -8,11 +8,7 @@ license: '[CC BY-ND 3.0](http://creativecommons.org/licenses/by-nd/3.0/us/)'
 alias: ['networking/configuring-static-ip-interfaces/']
 modified: Saturday, December 27th, 2014
 modified_by:
-<<<<<<< HEAD
-  name: Dave Russell Jr
-=======
   name: Dave Russell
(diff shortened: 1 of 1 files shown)


docs/networking/linux-static-ip-configuration.md
```diff
@@ -200,19 +200,19 @@ You must create the `/etc/sysconfig/network-scripts/ifcfg-eth0` file.
...
 
   # Adding a second public IP address.
-  IPADDR1=34.56.78.90
-  PREFIX1=24
+  IPADDR=34.56.78.90
+  PREFIX=24
 
   # Adding a private IP address.
-  IPADDR2=192.168.133.234
-  PREFIX2=17
```
buhman: This doesn't create aliases.

If non-unique names are used, only the last-defined `IPADDR`, `PREFIX` will have any effect.

buhman:
:thumbsup: after the changes to the CentOS section are reverted.

drussell393:
Reverted. Thanks Buhman <3

PR #147, opened 2015-01-06
Install MariaDB Galera Cluster with HAproxy and KeepAlived on CentOS 6.5

ShahriyarR:
While searching, i have found out that all related topics have some lack of information.
I decide to combine all necessary information into one complete tutorial.
Related firewall settings, configuration settings and testing tips are included.


alexfornuto:
Hello @ShahriyarR,

Wow, that sure is a big guide! In my initial scanning through, I noticed a few things I'd like to see corrected before taking a more serious look:

 - The English in the introduction, and throughout, is poorly structured and could stand some revisiting. If need be, we can help you with that.
 - Our minimum Linode plan is a 1GB RAM package, not 512.
 - Could the whole process be simplified by using the Linode [NodeBalancer](https://www.linode.com/nodebalancers) instead of configuring two extra Linodes for HAProxy?
 - How different of a guide would this be if it was compatible with CentOS 7?

Overall I'm very interested in this subject, and I hope we can make a great guide together!

ShahriyarR:
Hello @alexfornuto, 
As  you suggest:
> The English in the introduction, and throughout, is poorly structured and could stand some revisiting.  If need be, we can help you with that.

Now introduction is rewrited.

> Our minimum Linode plan is a 1GB RAM package, not 512.

Also changed Hardware specification, from at 512 MB to at least 1 GB.

> Could the whole process be simplified by using the Linode NodeBalancer instead of configuring two  extra Linodes for HAProxy?

For now i can't tell that, replacing HAproxy to Linode Loadbalancer will simplify works, but it could be tested and we can write(prepare) a new version of tutorial.

> How different of a guide would this be if it was compatible with CentOS 7?

The full process should be tested separately for CentOS 7, but i it is obvious that, the configuration files will be same, but firewall setting commands and some other terminal commands , such as restarting services etc. will change. 
Nevertheless all settings must be tested carefully, for introducing really working content to world.

PR #149, opened 2015-01-09
Correct php5-fpm listening socket

sjparsons:
The suggested nginx.conf should point to the unix socket that php5-fpm is already configured to be listening on.


alexfornuto:
Great, thanks!

PR #150, opened 2015-01-11
Remove cryptdevice parameter

rianhunter:
Only Arch Linux uses the `cryptdevice` parameter to dynamically specify where the encrypted disk resides. In Debian this is done statically through `update-initramfs`.


buhman:
:thumbsup: `cryptdevice` was never appropriate here.

PR #152, opened 2015-01-12
Update lemp-server-on-centos-7-with-fastcgi.md

otjeremy:
1. Current EPEL URL returns 404. Updated to reflect: https://fedoraproject.org/wiki/EPEL
2. Missing semi-colons cause nginx to fail on restart


buhman:
:thumbsup: LGTM

PR #153, opened 2015-01-12
Hadoop Clusters with Ambari

alexfornuto:


thetechstewart:
Reviewed and sent some additional requests back to the author, awaiting further input.

PR #156, opened 2015-01-13
Update squid-http-proxy-ubuntu-12-04.md

dev0da:
Fixed the wrong path which caused authentication not working.


alexfornuto:
:+1:

PR #167, opened 2015-01-22
Register DDNS addresses in Linode DNS

alexfornuto:


buhman:
:-1: 

* [`domain.resource.update`](https://www.linode.com/api/dns/domain.resource.update)'s `[remote_addr]` behavior is already documented
* "randomly run this task every $FOO hours" is the wrong way to do DNS/DHCP updates

brittonv:
If running $FOO is the wrong way, what is the right way?  

All ears if you have a better way!

alexfornuto:
@buhman constructive criticism is always appreciated. Do you have any suggestions for @brittonv on how to better implement this?

PR #169, opened 2015-01-26
Securing your DNS server from DDoS

alexfornuto:


buhman:
> iptables to successfully mitigate DDoS attacks against a DNS server

Hilarious.

:-1:; the only useful information from this TL;DR is "amplification attacks exist"--I suggest making that part of an article that talks about playing with bind in general.

ryanwohara:
This guide is very misleading and should be removed.

alexfornuto:
Thanks for the input @buhman and @ryanwohara. While I would very much like a guide or three on DDoS prevention and/or mitigation, it sounds like there are too many issues with this guide to make it a viable solution.

PR #170, opened 2015-01-27
Update running-a-mail-server.md

tyleregeto:
Remove Outlook.com as an External Mail Service, it is no longer supported. Microsoft has migrated the service over to Office 365, but it is a paid service. I have not updated the article to mention Office 365 because I'm not familiar with it. Just wanted to correct inaccuracy because it lead me astray.


alexfornuto:
Thanks! I'll add a link to 365.

PR #174, opened 2015-02-01
Update websites-with-nginx-on-ubuntu-12-04-lts-precise-pangolin.md

j-n-f:
a small note, this was something that tripped me up for a bit


alexfornuto:
I'm confused about the need for this note. If you follow the steps above you should be in `/opt/` when you untar nginx, which should make the `/opt/nginx-$VERSION` dir. Furthermore, future versions of this guide probably will not include directions on installing from source at all, since nginx now hosts it's own deb repositories for the newest versions.

PR #177, opened 2015-02-06
Update manage-web-content-with-wordpress.md

oppih:
Or it will not work because you cannot find wp-admin/install.php in the public_html


thetechstewart:
Great catch!

alexfornuto:
@oppih Thanks!

PR #179, opened 2015-02-06
How to install and configure OSSEC HIDS on Debian 7

thetechstewart:


threadswarm:
I think a guide on this topic would be a valuable addition to the Linode documentation library.

PR #181, opened 2015-02-11
Update hosting-a-website.md

rwky:
Reduced MaxClients for a 1GB node, 80 is way too high for the most common use of mpm_prefork which is a LAMP stack. 8 will mean that if each process uses 64MB RAM (not uncommon for PHP and can be quite low) that it'll use 512MB leaving the other 512MB for the OS/MySQL.

### DIFF
--- a/docs/websites/hosting-a-website.md
+++ b/docs/websites/hosting-a-website.md
@@ -68,7 +68,7 @@ Installing Apache is easy, but if you leave it running with the default settings
 	StartServers 2
 	MinSpareServers 6
 	MaxSpareServers 12
-	MaxClients 80
+	MaxClients 8
 	MaxRequestsPerChild 3000
 	</IfModule>
 	~~~


drussell393:
I feel like 8 would be going way too low here. Working out the MaxClients setting is usually done by taking the amount of RAM, subtracting the amount for MySQL/OS and then dividing it by the average size of your Apache process. Now, if each process is taking up 64MB of RAM, it sounds like you need to optimize your application.

If PHP is taking up 64MB of RAM for each request, your application is doing way too much and you should definitely look into optimizing it. I run a fairly large website for one of my clients, and their average Apache process size is just under 15MB. 

Even if we assume 15MB is the average size for everyone's website, the MaxClients could be 30 and give the OS/MySQL 512MB. Setting the MaxClients setting to 8 will cause extremely slow load times on most websites anyways, especially if they get a decent amount of views, and it's even possible that it'll cause the website not to load at all because the MaxClients directive will be set too low. 

In most cases, Apache's default of 150 is sufficient. Bottomline, if your application isn't optimized, optimize your application. I would advise against changing your MaxClients directive to cover up the fact PHP is using up way too much memory per request.

rwky:
15MB isn't anywhere near enough to run common applications such as Wordpress. Check the Linode forums, people have been using the guide and their servers OOM.

alexfornuto:
@ellekrout is currently updating our resource testing and tuning guides, so she has been assigned to update this guide as well with settings more appropriate to a Linode 1GB based on newer tests.

ellekrout:
This was addressed internally and we've changed the MaxClients amount to 30 in the appropriate guides:
https://www.linode.com/docs/websites/hosting-a-website/
https://www.linode.com/docs/websites/lamp/lamp-server-on-debian-7-wheezy
https://www.linode.com/docs/websites/lamp/lamp-server-on-ubuntu-14-04

PR #187, opened 2015-02-18
Protecting your Linode with TCP Wrappers

thetechstewart:


AliceWonderMiscreations:
`TCP wrappers offer less functionality that a full-blown firewall'

Looks like a typo (the kind I frequently make) - the `that' should be a `than' - line 9.

francismcnamee:
Sorry about that!

PR #188, opened 2015-02-18
Webmin on CentOS

thetechstewart:

### DIFF
--- a/docs/websites/cms/webmin-centos7.md
+++ b/docs/websites/cms/webmin-centos7.md
@@ -1,0 +1,89 @@
+iInstall Webmin on CentOS using the tar.gz file
+
+Introduction
+
+Webmin is a system that allows the management of the main servers (Unix platform) as; Apache, DNS, Squid, DHCP, file sharing, MySQL database among many others.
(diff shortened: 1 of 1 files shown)


ryanwohara:
Can we close this? There's no reason Linode should maintain their own documentation when Webmin already does it:

http://www.webmin.com/rpm.html

docs/websites/cms/webmin-centos7.md
```diff
@@ -0,0 +1,89 @@
...
+
+To install http server run the following command in the shell;
+
+
+yum install httpd -y
+
+
+Soon after install Perl
+
+yum install perl perl5 -y
```
alexfornuto:
```
# yum install perl perl5 -y
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
 * base: mirrors.linode.com
 * extras: mirrors.linode.com
 * updates: mirrors.linode.com
Package 4:perl-5.16.3-283.el7.x86_64 already installed and latest version
No package perl5 available.
Nothing to do
```

alexfornuto:
@ryanwohara I tend to agree with you, unless we get a guide that details better how to install the software for our platform better than the official documentation. However, I don't feel that this guide falls into that category.

Webmin offers a yum repository to install and update the software from. This is a preferred method to installing from a tar ball, since the software can more easily be kept up-to-date.

PR #191, opened 2015-02-24
Update lamp-server-on-centos-7.md

stonemirror:
The "register_globals" keyword is deprecated as of v5.4.0, see http://php.net/manual/en/ini.core.php#ini.register-globals


alexfornuto:
:+1:

PR #194, opened 2015-03-01
Fix mysql link reboot survival

drussell393:
This is to fix the link in the Reboot Survival Guide.


alexfornuto:
Is the link in the reboot survival guide wrong? If so, is it not better to fix that link itself?

thetechstewart:
Looking into fixing this link now - it should be directed to the current link, not pointed to a redirect.

thetechstewart:
Link repaired in the original document.  @drussell393, feel free to close this PR if everything looks good on your end.

drussell393:
@alexfornuto Ultimately, yes, but I wasn't sure if that link was used elsewhere also -- which is why I just set up the redirect to fix $all instances rather than "Let's assume this is the only instance ever." 

But, if you guys are content with fixing this one instance, and you're confident this is the only instance where this link is used, then feel free to close this PR :)

thetechstewart:
I neglected a slash at the beginning of that link, which is what broke it in the first place.  That should only affect this link.  Closing PR.

PR #195, opened 2015-03-02
need to replace mpm_event_module with mpm_prefork_module

imdecoder:
only comment out pm_event_module will fail to start httpd.service, and give `httpd: Configuration error: No MPM Loaded.`, need replace with mpm_prefork_module


alexfornuto:
While the tech there sounds right (need to verify) the sentence structure in your proposed edits needs to be fixed.

alexfornuto:
Thanks for being so patient!

PR #198, opened 2015-03-13
Convert root filesystem to BTRFS

disconn3ct:
This is a fairly simple process that allows users to convert their root filesystems to use BTRFS. It has been tested repeatedly, and should fail non-destructively in the event that something goes wrong. Although it is written for Arch Linux in a configuration close to standard, it is adaptable enough to be used on different distributions or with more unique environments.

All feedback and criticism welcome.

Thanks!


disconn3ct:
No comments yet? Its been in for over 2 months..

alexfornuto:
@disconn3ct I actually tested this last week and it works fine! Later this week I'll be testing it on Debian next.

alexfornuto:
Since Debian doesn't have a `mkinitcpio.conf`, I'm not going to spend more time looking into making this Debian compliant. I've created https://github.com/disconn3ct/docs/pull/1 which formats your guide to match our docs.

disconn3ct:
It looks like the config would live in /etc/initramfs-tools/initramfs.conf but I'm not sure on the specific details of the changes.

alexfornuto:
Please replace any images of text output with the text output itself as an unformatted text block.

alexfornuto:
I'm closing this PR due to inactivity. @disconn3ct Please let us know if you are willing to make the final changes requested and we can reopen.

PR #199, opened 2015-03-16
Django on Ubuntu 14.04

thetechstewart:

### DIFF
--- a/docs/websites/frameworks/django-on-ubuntu-14-04.md
+++ b/docs/websites/frameworks/django-on-ubuntu-14-04.md
@@ -1,0 +1,154 @@
+## About Django
+
+Django is a great web application development framework for the perfectionist with deadlines built in python programming language. It provides a very good structure and common methods that help to do the heavy lifting when writing web applications.
+
+If you have never used django before and are looking for an installation guide on your **Ubuntu 14.04 LTS** then you are in the right place. In this article we will teach you how to setup a very simple django development environment step by step.
+
+## Some Tools We Need
+
+There are many different ways to install the django framework in Ubuntu. In order to stay updated with the latest python development industry standards we need to install a few tools before getting Django up and running on our machine.
+
+The first tool is called `pip` which is a very practical python package managment system widely used by python developers all over the world. I mainly use `pip` to install packages for my python projects, generate dependencies of a project and also uninstall different python packages from my machine when they are no longer required.
+
+Before installing `pip` we higly recommend to our users to update the package index on their machine with the help of the command shown below.
+
+    sudo apt-get update
+
+The following command can be used to install the `pip` python package manager inside `Ubuntu 14.04 LTS`.
+
+    sudo apt-get install python-pip
+
+Now we can easily install python packages on our `Ubuntu 14.04 LTS` machine by following the syntax shown below.
+
+    pip install name-of-python-package-here
+But we do not recommend installing packages yet! Python developers use another tool during the development of their python projects.
+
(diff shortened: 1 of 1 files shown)


jetbird:
Hi Sterge,

Thanks for the feedback. The article explains how to install django on ubuntu 14.04 LTS not how to setup a django development environment. I explained how to install django in a virtual environment as it is the best way to prevent conflict dependencies.

Thanks for the other information you provided. I agee with you, I love django :)

docs/websites/frameworks/django-on-ubuntu-14-04.md
```diff
@@ -0,0 +1,154 @@
...
+Now we can easily install python packages on our `Ubuntu 14.04 LTS` machine by following the syntax shown below.
+
+    pip install name-of-python-package-here
+But we do not recommend installing packages yet! Python developers use another tool during the development of their python projects.
+
+This tool is called `virtualenv`. It is really useful as it helps to manage dependency conflicts of different python projects on the same machine. In short words the tool `virtualenv` helps to create isolated virtual environments where you can develop without any worries about package conflicts.
+
+To install `virtualenv` on `Ubuntu 14.04 LTS` the following command can be used.
+
+    sudo pip install vitualenv
```
alexfornuto: Why would I get this output the first time I attempt to use `pip`?

```
sudo pip install vitualenv
Downloading/unpacking vitualenv
  Could not find any downloads that satisfy the requirement vitualenv
Cleaning up...
No distributions at all found for vitualenv
Storing debug log for failure in /home/alex/.pip/pip.log
```

alexfornuto:
Here are some notes on this guide, as I am testing it.

It's very common for people to preface every command with something like "the following command will..." or "do so with the following command". You can assume that the reader knows that a command is a command. You can shorten the step to "Install XYZ" or "move the file", etc. 

Technical documents should be objective. It's better to explain why something is good than say it is just "very good".

If you have multiple steps, number them.

If the guide is on X, explain X. If X depends on Y, only go into as much detail on Y as you need to get X going. General use case descriptions and examples of Y belong in a separate guide on Y, which can be linked in the guide on X.

Expect the user to run every command you list. If you instruct them to go into a virtual environment, be sure to have them leave it before performing more system-wide commands. All steps should be linear.

If certain commands require root privileges, be consistent throughout the guide. Either use `sudo` whenever required, or preface at the start of the guide that the root user is assumed.

After finishing this document, it seems to me that this is not a guide on Django. It is a guide on the python virtualenv tool, with Django used as the example end goal. If you would like to resubmit this guide as a virtualenv Tutorial, feel free, but as it doesn't go into any detail on Django itself we cannot accept the guide as is.

jetbird:
What do you mean it is not a guide on Django? I did not say it is not a
guide on Django. The title is How To Install Django On Ubuntu 14.04 LTS.

There are many ways to install Django on Ubuntu and the objective of this
guide is to teach people how python developers install it on their Ubuntu
machines.

If you want I can resubmit this guide as virtualenv Tutorial. Do you want
me to remove the django part?

jetbird:
So the thing is to teach the readers how to install django like a
professional developer. If they know about virtualenv than they don't have
to read the guide at all. The reason why I have explained virtualenv is for
the pople that do not know about it.

So in this guide  they learn about virtualenv, they setupa virtual
environment and in the end they have django installed on their machines. I
do not know what your background  is with python so I hope you understand.

alexfornuto:
A reader who sees this title would expect the guide to focus on Django. This guide focuses on Python virtual environments, and only gets to Django as a last step, with the single line to install it through pip. 

A guide on how to install Django, or any software, should be about the software itself. Not just the act of installing, which is usually trivial thanks to package managers, but how to configure and begin working with that software. You are of the opinion that Django is best installed inside of a python virtual environment. That could very well be a suggestion in the type of guide I've just described, but it should not be the full contents.

PR #202, opened 2015-03-29
Added note to ensure pre-requisites are installed before trying to build...

koipondinteractive:
... NVM from source.


drussell393:
Can you clarify which prerequisites you're referring to here? The shell script should take care of most of them, if not all.

alexfornuto:
Closing for lack or a response from the author to @drussell393's questions.

PR #203, opened 2015-03-31
Mention commands need to be prefixed with label

Odaeus:
I couldn't see it mentioned anywhere in the guide that you need to write the name of the Linode before each command. I've added a brief example.

### DIFF
--- a/docs/networking/using-the-linode-shell-lish.md
+++ b/docs/networking/using-the-linode-shell-lish.md
@@ -153,7 +153,7 @@ These are the fingerprints for the Lish gateway in our London data center (lish-
 Understanding Lish Commands
 ---------------------------
 
-The Lish shell provides access to many functions which are otherwise only accessible via the Linode Manager web-based administration tool. Enter the `help` command to see a full list of available commands. The output provides an introduction to Lish functionality:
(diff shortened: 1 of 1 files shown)


FelicianoTech:
I agree that this is confusing but it looks like this specific PR may not be the exact solution. Those commands work once you choose a Linode.

From the Lish gateway, you can type the label of your Linode, which brings you to that Linode's console. If you then control-D out of that, you get to lish on the host which allows you to run those commands without any prefix.

Odaeus:
Thanks @FelicianoTech, feel free to amend it. Though I think prefixing the label is less work than activating the console and then escaping from it as you suggest. I didn't know about that method either!

alexfornuto:
@Odaeus @FelicianoTech I've made https://github.com/linode/docs/pull/232, which I think addresses both of your concerns, as well as some other formatting changes I felt improved the guide. What do you both think?

FelicianoTech:
I commented in the other PR however basically, I like it.

alexfornuto:
I merged #232 so I'm closing this PR. Thanks for bringing this to our attention @Odaeus!

PR #206, opened 2015-04-05
DNSSEC with NSD

AliceWonderMiscreations:
This pull request contains two files: markdown documentation and an image file.

I would like the author name listed as Alice Wonder

If accepted, linode credit is preferred, I'm not sure what my linode account number is, I could not find it in my account tab. My Linode login username is FunkyRes and my "real" name is Michael A. Peters.


AliceWonderMiscreations:
It looks like many linode docs are http://creativecommons.org/licenses/by-nd/3.0/us/
That is fine by me if this is accepted.

drussell393:
Michael,

I am going to be completely honest and uncompromisingly forthright, this guide is way too long. Now, while I agree there are useful bits of information in the guide, I seriously think that you can considerably compress the documentation by removing your personal opinions. Having read the entire documentation, I lost determination about halfway through.

Most of the documentation is cluttered with "I" or "me" or "my," words that should, in my opinion, never be included in a documentation. A documentation should be straightforward, to the point, and provide only factual information. Saying things like, "That problem has existed for me since at least the CentOS 5 days and that might be why it is not packaged for CentOS / RHEL in the official RPM repositories," provide no benefit to the guide itself. It adds extra content, but the reader gets no other benefit than your opinion and struggles since CentOS 5.

Specific statements that caught my attention for different reasons:

> "This document was written with CentOS 7 in mind but it should work with most Linux distributions with very little tweaking needed."

The configurations will likely be the same, yes, but the installation process will be altogether different. I would recommend keeping this guide as the title suggests, CentOS 7, and not mentioning any other distributions. This will just lead to confusion, and likely people will ask the feature question of, "How do I tweak it for Ubuntu?" which may not be as easy as it seems.

> "I need to warn you that migrating to DNSSEC is not something where you can just set it up and forget it for awhile. It does require frequent maintenance to roll over one of the keys. You can automate that, but that automation is beyond the scope of this document."

While it may be prudent to tell the reader that their setup will need constant maintenance, there's no mention of how they'd go about doing this maintenance. Additionally, you mention automation, but give no assistance in automating the process. This just adds content to the documentation with no actual benefit.

> "I have however attempted to make the key rollover a simple matter of running a few scripts."

Again, this provides personal experience. This doesn't benefit the reader at all, as they don't have the scripts. I believe you may have included the scripts further down, but if that's the case you should mention the automation and constant maintenance in that section. The introduction becomes hazy with these statements.

> " I personally find it easier to configure than bind, it does what an authoritative name server needs to do and it does it quite well." 

Once again, this is about personal opinion and likely provides no benefit. 

> "I personally prefer not to do that, I do not like to use domain names for my nameserver that are part of the zones they control. I prefer to register a separate domain that is only used for my nameservers and host the zone for that domain on a third party nameserver, such as the Linode provided nameserver. It is simpler and simplicity has its benefits."

Once again, this is about personal opinion and likely provides no benefit. 

> "I suggest using a .com or .net TLD for the nameserver domain, it allows you to use http://www.internic.net/whois.html to easily verify that your nameservers are in fact seen as nameservers."

While this is beneficial, it's probably not necessary to mention this. You can WhoIs any TLD, and at the end of the day, the reader is going to use their domain (regardless of the TLD) to set up the nameservers.

> "Once Linode's DNS server has updated (usually less than 15 minutes after you create the zone)"

This isn't actually accurate. The DNS Manager DNS servers will update between 15 minutes and one hour of the DNS record being entered or changed.

I'm not going to keep going through statements to point them all out, I think you'll understand the gist.

As a side note, it would be better if your documentation complied with [RFC 5737](http://www.rfc-base.org/txt/rfc-5737.txt) and [RFC 3849](https://tools.ietf.org/html/rfc3849). These RFCs define documentation subnets for IPv4 and IPv6 addresses. This will prevent confusion throughout the IP format, and it's all around better to comply with the RFCs.

You may want to consider making this a series of documentations. The length of the document is far too much, and most people will likely get intimidated at first glance. I would recommend breaking up the introductory parts, configuration parts, and automation parts into their own documentations. The introduction alone (defining of the concept, etc) is plenty for its own documentation.

You could put it in ../docs/networking/dns/dnssec or similar. However, I really believe that the document will be considerably compressed with the subtraction of personal opinion. Most of the documentation is currently cluttered with "I" "me" or "my" and personal experiences. Ideally, you'll want to cut those out of the documentation and leave it to pure facts, how to set it up, and don't elude to things you won't discuss in the documentation. The idea of the documentation isn't to provide a basis on what your experiences are, or what can be done, but rather how to do what can be done.

-- Dave Russell Jr

AliceWonderMiscreations:
Yeah I could fix some of that stuff, but looking at the time I already put into it I am already at less than minimum wage, I'm just going to cut my losses, linode docs clearly isn't for me.

PR #207, opened 2015-04-06
Updating Debian 7

thetechstewart:


gacanepa:
@ravenx99 Note that the intent of this brief guide is not upgrading to Jessie, but updating from a Wheezy release to a more recent one, thus the use of aptitude instead of apt-get, which as you say it's the recommended tool by the Project for upgrades to a newer version of the distribution.
That being said, and since you brought up an interesting point, can you think of anything that could go wrong in this process? I would appreciate your input in order to improve the guide and make it an interesting topic for publication.

alexfornuto:
@gacanepa I was hoping to see some more discussion from the community, but I'm in agreement with @ravenx99. This guide is too short and written on a subject that is already well documented by the original project maintainer.

gacanepa:
@alexfornuto I agree. This guide was meant to be a sample guide and I didn't know it would be independently reviewed by the community - but I actually like the idea! Now that I am more familiar with Linode's standards for articles I believe I will be able to submit another guide that complies with them. Thank you for getting back to me!

PR #208, opened 2015-04-07
Update native-ipv6-networking.md

kdhartstrom:
Correct service command is "service network restart" not _networking_


alexfornuto:
Thanks!

PR #210, opened 2015-04-09
Clarify some unanswered questions about linode images

ethanpil:


ethanpil:
Per an email request with support on 4/9/2015

docs/platform/linode-images.md
```diff
@@ -13,7 +13,10 @@ published: 'Thursday, September 25th, 2014'
...
+Linode Images allows you to take snapshots of your disks, and then deploy them to any Linode under your account. This can be useful for bootstrapping a master image for a large deployment, or retaining a disk for a configuration that you may not need running, but wish to return to in the future. Linode Images will be retained whether or not you have an active Linode on your account, which also makes them useful for long term storage of a private template that you may need in the future. There is no additional charge to store images for Linode users, with a limit of 10 images per account.
+
+ {: .note }
+>If you delete an Linode, the IP address will be removed. So if you redeployed an image, you'd get a new IP address.
```
alexfornuto: I would change this line to:
`When you delete a Linode, its IP address(es) are removed. So if you deploy a new Linode from an image, the Linode will have a new IP address assigned to it.`

alexfornuto:
Thanks! This is a useful addition. I made one note on line 19. Once that's adjusted we can merge.

docs/platform/linode-images.md
FelicianoTech: While that statement is technically true, it's completely outside of the realm of an image. An IP address would never be in an image on any platform.

The IP can change, the MAC can change, FQDN can change, resources, etc. Much more than simply an IP address.

alexfornuto: Good point @FelicianoTech. Maybe instead of talking about IP addresses at all, we should set straight early in the guide that an Image is of the disk itself, and carries no other aspects of the Linode it was created from. How do we all feel about that?

ethanpil: I included the line about IP addresses so that the end user will remember that an image wont necessarily work as expected if they assumed the IP address was the same. For novice Linode users I think it might be easy to assume that they can just image and restore.

alexfornuto:
@ellekrout is going to see if she can come up with another way to express the issue discussed on L19

docs/platform/linode-images.md
ellekrout: My attempt:

When saving a Linode image it is the aspects of the Linode that are on the **disk image** that are saved, not any additional aspects such as IP addresses, fully qualified domain names, and MAC addresses.

Not in love with it, but it's a starting point?

alexfornuto: Also, the current limits on Linode images are:

1. 2GB per Image
2. 10GB total image storage
3. 100 Images per account.

alexfornuto:
@ethanpil we went ahead and brought your changes in with ours in https://github.com/linode/docs/pull/233. Thanks for bringing it to our attention!

PR #225, opened 2015-05-13
Update custom-compiled-kernel-with-pvgrub-debian-7-ubuntu.md

shinji257:
Add bc to the apt-get install line.  Needed during the main make step.  Doesn't seem to always be present in the OS install.

### DIFF
--- a/docs/tools-reference/custom-kernels-distros/custom-compiled-kernel-with-pvgrub-debian-7-ubuntu.md
+++ b/docs/tools-reference/custom-kernels-distros/custom-compiled-kernel-with-pvgrub-debian-7-ubuntu.md
@@ -24,7 +24,7 @@ Update your package repositories and installed packages, install the development
 
     apt-get update
     apt-get upgrade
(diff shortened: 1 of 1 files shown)


alexfornuto:
@levlaz Maybe we should discuss making a point of ensuring that `bc` is in all future distro images. Or is in uncommonly used enough that we'd rather just install it when needed?

levlaz:
This is very odd. bc should be installed by default in debian and ubuntu. It is for a fact in debian 7, have not checked other. @shinji257 which distro are you running where bc is not installed.

shinji257:
I'm running Ubuntu 14.04.  At the time it was the image supplied by Linode.  I have not tried a new deploy to see if this is still the case or not.

levlaz:
@shinji257 Thanks! This seems like more of an issue with the distro build as @alexfornuto mentioned than with the guide. We need to make sure that this is available in the distribution itself.

alexfornuto:
@thetechstewart will make sure that this package is in future builds as needed.

PR #226, opened 2015-05-14
Add pv-grub distro kernel instructions for Debian 8 (Jessie).

thehunmonkgroup:
Followed the instructions for Wheezy, and they worked for me on Jessie.


alexfornuto:
Good thinking! I think we'd prefer to just update the Debian 7 line to something like:

```
 Debian 7 (Wheezy) / Debian 8 (Jessie)
```

thehunmonkgroup:
So executed...

PR #229, opened 2015-05-19
Java Development with WildFly on CentOS 7

thetechstewart:

### DIFF
--- a/docs/applications/development/wildfly.md
+++ b/docs/applications/development/wildfly.md
@@ -1,0 +1,436 @@
+Introduction
+============
+JBoss AS (Renamed to [WildFly](http://wildfly.org/)) has been used for many years as one of the best known performing & free community Java Application Server. The attention even increased when Red Hat started providing commercial support to its enterprise version [JBoss EAP](http://www.redhat.com/en/technologies/jboss-middleware/application-platform) with support to on-premise, virtual, or cloud. Red Hat also provides commercial support for Red Hat Enterprise Linux & building huge ecosystem of multiple products to serve its clients.
+
+Many software companies working with Java technology are targeting the stack of WildFly on CentOS so it supports clients with limited budget, and also supports JBoss EAP on Redhat Enterprise Linux for customers looking for commercial support, so they ensure their software meets various customer segments.
+
+This article has been introduced to show one of the most common architecture in Java application hosting stack, it consists of:
+
+ 1. CentOS 7: As Operating System.
+ 2. MySQL 5.6.24 Community Server: As RDBMS.
+ 3. WildFly 8.2 Final: As Java Application Server.
+ 4. Apache HTTP 2.4.6: As HTTP Server for serving static content & using mod_jk 1.2.40 for directing calls to WildFly (Apache as reverse proxy).
+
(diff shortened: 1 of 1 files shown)


ellekrout:
Hello @ashraffouad! Sorry for taking so long to get back to you. I've done a tech test and have some notes:

- For the MySQL installation prereqs I would link to the [LAMP on CentOS 7](https://www.linode.com/docs/websites/lamp/lamp-server-on-centos-7) guide instead of Hosting a Website, which is for Debian/Ubuntu.

- When running the wildfly-install.sh script it hits a hiccup during the `systemctl enable $WILDFLY_SERVICE` line, and outputs the following:

> wildfly.service is not a native service, redirecting to /sbin/chkconfig.
> Executing /sbin/chkconfig wildfly on
> The unit files have no [Install] section. They are not meant to be enabled
> using systemctl.
> Possible reasons for having this kind of units are:
> 1) A unit may be statically enabled by being symlinked from another unit's
>    .wants/ or .requires/ directory.
> 2) A unit's purpose may be to act as a helper for some other unit which has
>    a requirement dependency on it.
> 3) A unit may be started when needed via activation (socket, path, timer,
>    D-Bus, udev, scripted systemctl call, ...).

- When I reach the step in which you add the Datasource it failed on both servers I tested it on with this error:

>Unexpected HTTP response: 500
>
>Request
>{
>    "address" => [
>        ("subsystem" => "datasources"),
>        ("data-source" => "MySQLDS")
>    ],
>    "operation" => "test-connection-in-pool"
>}
>
>Response
>
>Internal Server Error
>{
>    "outcome" => "failed",
>    "failure-description" => "JBAS010440: failed to invoke operation: JBAS010447: Connection is not >valid",
>    "rolled-back" => true
>}

ellekrout:
Since a month has passed with no author response to the above concerns, closing this PR.

ashraffouad:
Sorry @ellekrout I had a long vacation, and couldn't access server, sorry for late reply. Below reply for your comments:

1- Link for [LAMP on CentOS 7 guide](https://www.linode.com/docs/websites/lamp/lamp-server-on-centos-7) is mainly for mariadb, I'm afraid it will cause confusion if someone doesn't know mariadb is like mysql, I think we can either change as you wish or I can write the complete steps in this command, at the end they are the same commands executed but we will replace mariadb with mysql.

2- Adding WildFly as a service, I added a fork (commit: e56767abdecbc978172edd4b85486f03539f0122) to the file to fix this issue, mainly in the script /opt/wildfly-install.sh

3- I could not reproduce this error, can you please tell me exactly where it is occurring in the steps?

Sorry, I'm new to Github, and afraid if my changes/fork is not shown, please let me know if I should do more steps to show you my commit.

alexfornuto:
@ashraffouad If you're interested in pursuing this guide and you already have a GitHub account with updates, it's probably best for you to make your own Pull Request against Linode/docs. That way your changes automatically update the PR while it's staged, and we can work more directly with you, taking advantage of inline comments as well as the rest of the community. If you reference this PR in your own, using the notation `#229`, it will create reference links on both pull requests, so we can follow the historical conversation.

alexfornuto:
Closing so we can continue on #260.

PR #230, opened 2015-05-20
High Performance Nginx and PHP with Debian 8

thetechstewart:

### DIFF
--- a/docs/websites/nginx/nginx-php-fpm-debian-8.md
+++ b/docs/websites/nginx/nginx-php-fpm-debian-8.md
@@ -1,4 +1,4 @@
-High performance Nginx and PHP over Debian 8
+ performance Nginx and PHP over Debian 8
 19th May 2015 by Javier Briz
 
 
@@ -20,39 +20,53 @@ Run the following as root to install nginx, PHP and MySQL connector:
 
 Make sure nginx and PHP-FPM are running
 
-        ps aux|grep -e nginx -e php
+        systemctl status nginx
+        systemctl status php5-fpm
         
 #Nginx setup
 Now we have to setup nginx server to relay on PHP-FPM for PHP requests. 
 By default, PHP-FPM on debian listens on a socket. We are putting everything in one single server so this is ok.
 To find the path for the socket run:
 
-        cat /etc/php5/fpm/pool.d/www.conf |grep 'listen ='
+        grep 'listen =' /etc/php5/fpm/pool.d/www.conf
 
 Probably, it is located in `/var/run/php5-fpm.sock`
 
 We will tell nginx where is this socket in a *server* context in the nginx setup.
-In other words, add following lines to `/etc/nginx/sites-enabled/default`, below start of *server* section:
+It is done by modifying your default host file content. Edit `/etc/nginx/sites-enabled/default` and replace the default content with this:
 
-        location ~ [^/]\.php(/|$) {
-          fastcgi_param REQUEST_METHOD $request_method;
-          fastcgi_pass unix:/var/run/php5-fpm.sock;
-          fastcgi_index index.php;
-          include fastcgi_params;
+        server {
+                server_name _;
+                listen 80 default_server;
+                listen [::]:80 default_server;
+                
+                root /var/www/html;
+                index index.php index.html index.htm;
+
+                location / {
+                        try_files $uri $uri/ =404;
+                }
+
+                location ~ \.php$ {
+                        try_files $uri =404;
+                        fastcgi_pass unix:/var/run/php5-fpm.sock;
+                        fastcgi_index index.php;
+                        fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
+                        include fastcgi_params;
+                }
         }
 
 Restart nginx by running:
 
-        /etc/init.d/nginx reload
+        systemctl restart nginx
         
 #Test
 Done! Let's test the web server. Just write the following content in /var/www/html/index.php:
 
         <?php
         phpinfo();
-        ?>
 
-And now go to http://example.com/index.php and, if everything is ok, you will see the phpinfo for the installed version.
+And now go to http://[your_linode_node_ip]/index.php and, if everything is ok, you will see the phpinfo for the installed version.
       
 {: .note}
 >
@@ -143,7 +157,7 @@ This is not always possible since some pages may allocate higher amounts of memo
 
 After modifying PHP-FPM setup, just restart it.
 
-        /etc/init.d/php5-fpm restart
+        systemctl restart php5-fpm
 
 #Next steps
 
@@ -162,4 +176,5 @@ A few good tools to do this are iotop, htop, zabbix (for the long term), ps, fre
 - `zabbix` is (imho) the best open source monitoring tool, altough it is a bit complex to install. You can get more information [on zabbix.com](http://www.zabbix.com/).
 
 Other tools, like `ps`, `free` and much others, are usefull when you need a particular data, because its performance is higher than that of the previous tools, but will not offer you a more complete overview of your system.
+
 
--- a/docs/websites/nginx/nginx-php-fpm-debian-8.md
+++ b/docs/websites/nginx/nginx-php-fpm-debian-8.md
@@ -1,3 +1,4 @@
+http://markdownlivepreview.com/
 
 High performance nginx and php over Debian 8
 19th May 2015 by Javier Briz
@@ -9,9 +10,11 @@ able to connect to a MySQL server, but the MySQL server installation itself will
 there are already [lots of guides](https://www.linode.com/docs/databases/mysql/) on this topic. 
 
 #Installation Prerequisites
+First of all, follow the [Getting Started Guide](https://www.linode.com/docs/getting-started/).
+It is also usefull to have a Fully Qualified Domain Name to have a handy pointer to the server.
+will be usefull 
 You will need to add non-free to your `sources.list`. Unfortunately, we'll need to download php-fpm from non-free.
 
-        # cat /etc/apt/sources.list
         deb http://ftp.debian.org/debian/ jessie main contrib non-free
         deb http://security.debian.org/ jessie/updates main
 
@@ -51,9 +54,8 @@ Restart nginx by running:
         /etc/init.d/nginx reload
         
 #Test
-Done! Let's test the web server. Just write the following content in /var/www/html/index.php
+Done! Let's test the web server. Just write the following content in /var/www/html/index.php:
 
-        # cat /var/www/html/index.php
         <?php
         phpinfo();
         ?>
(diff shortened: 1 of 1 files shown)


sanvila:
*need to download php-fpm from non-free*

WTF??

algspd:
You will need fastcgi, and it is in non-free section of debian repos:
https://packages.debian.org/jessie/libapache2-mod-fastcgi

Maybe a bit confusing, I did not refer to php5-fpm package

sanvila:
WTF again.

That's for Apache and the guide says "Nginx".

In fact, Debian 8 has Apache >= 2.4, so even if you are using Apache, you don't need fastcgi anymore.

algspd:
You are totally right. I just sent the changes to be merged.
Thanks!

thetechstewart:
@sanvila - Thank you for your comments!  In the future, please provide clarification on your original comment so that authors can address them promptly.

sanvila:
Sorry to be terse, I was just assuming it would be pretty obvious (as we say in Spain: "A good listener needs few words").

alexfornuto:
All the same, "WTF" is not a constructive comment.

sanvila:
Some comments:

* There is no need to define `SCRIPT_FILENAME` if `fastcgi.conf` is used instead. Try:

        diff -ru /etc/nginx/fastcgi_params /etc/nginx/fastcgi.conf

* I would not write php-fpm or PHP-fpm but PHP-FPM instead. Reason [here](http://php-fpm.org/).
* Same for PHP and Nginx, Their capitalization should ideally be unified across the text.
* I would not write "myserver" but "example.com" instead. [RFC 2606](http://www.rfc-base.org/txt/rfc-2606.txt) explains why.
* I would write `service nginx reload` to reload a service, not `/etc/init.d/nginx reload` which is the old fashioned way (this also applies to a restart later in the text).

algspd:
>> There is no need to define SCRIPT_FILENAME if fastcgi.conf is used instead. Try:
>> diff -ru /etc/nginx/fastcgi_params /etc/nginx/fastcgi.conf
You are right, it's redundant as it is the default value.

>> I would not write php-fpm or PHP-fpm but PHP-FPM instead. Reason here.
>> Same for PHP and Nginx, Their capitalization should ideally be unified across the text.
I totally agree with this, it's just a a cosmetic measure but it is nicer to read a rightly capitalized text.

>> I would not write "myserver" but "example.com" instead. RFC 2606 explains why.
Yep

>> I would write service nginx reload to reload a service, not /etc/init.d/nginx reload which is the old fashioned way (this also applies to a restart later in the text).
service or systemctl? It depends on the service manager you are using, all three ways end up invoking the same script.

thetechstewart:
@algspd - in reference to your last response, your guide would need to reference the default init/service manager for the Linode provided image.  In the case of Debian 8, which has converted to SystemD for init, you would need to utilize systemctl commands to restart your services.

sanvila:
Actually, "service" is init-system-agnostic and will always work. There is no need to be systemd-specific. Even if Debian 8 has systemd by default, the user is free to install sysvinit and that's still a supported configuration.

Since I believe that a manual that will always work is better than a manual that will only work "almost always", I recommend "service".

thetechstewart:
@sanvila - While there are aliases present in Debian, and as one who prefers sysvinit I'd like it to stay that way, any guides published for Debian 8 will be modified to use equivalent SystemD commands during the editing process on our end.  If the author is able, we prefer that the guides be submitted with commands using the SystemD syntax whenever possible.

sanvila:
Ok, so it's not better, it's just that you *like* to be gratuitously sysvinit-incompatible. I fail to see the rationale for that, if there is one at all. If I were to write a guide, I would put the interest of the readers first, not my own preferences.

In either case: If you really want to follow systemd syntax, I would suggest that you start by spelling it correctly: it's systemd, not SystemD or anything like that:

http://www.freedesktop.org/wiki/Software/systemd/#spelling

thetechstewart:
@sanvila - We're putting the distribution defaults first, and basing this around the init system that the reader will be using, not the aliases to the older version.  As the init system has been replaced, and Debian shows no signs of switching back to sysvinit, our Debian guides will contain systemd syntax going forward.

I will ask again, as @alexfornuto noted previously, that you stick to constructive commenting.  This thread is for comments on the technical accuracy of the guides being reviewed, not for soapbox debate of various init systems, or non-sequitur arguments with Linode employees.

sanvila:
The only non-sequitur I see here is that Debian 8 somehow *implies* systemd. It does **not**.

Believe me that I try to be as constructive as I can, so if I have to explain it in detail, word for word, I will do:

    # cat /etc/debian_version 
    8.1
    # which service
    /usr/sbin/service
    # dpkg -S /usr/sbin/service
    sysvinit-utils: /usr/sbin/service
    # dpkg -s sysvinit-utils | grep Essential
    Essential: yes
    # dpkg -s systemd | grep Essential
    [no output]

I'm **all** for technical accuracy, and the guide says "Nginx and PHP-FPM with Debian 8". Debian prides itself to be an "Universal OS" and, as a result, a Debian 8 system **may** run without systemd. But it may **not** run without the "service" command as shown above.

So, if the guide says Debian 8 and not "Debian 8 with systemd" or "Debian 8 installed from a Linode image in which the admin didn't change any default", it will be **more** accurate to use "service".

In either case, your guide, your rules, so if you want to make "Debian 8" to imply "systemd" (when in fact it does not), do as you like, but then **please** don't talk about "technical accuracy", ok?

sanvila:
@algspd If you drop the line saying

   fastcgi_param SCRIPT_FILENAME

then you should use `fastcgi.conf`, not `fastcgi_params`.

algspd:
Could you be more specific? It is working perfectly without any more change.

sanvila:
@algspd Sure: I suggested two things to be done at the same time:

Using fastcgi.conf instead of fastcgi_params.
Dropping the SCRIPT_FILENAME line.

The end result should be the same because fastcgi.conf already includes the line being dropped.

But the last version I read just drops the SCRIPT_FILENAME line and keeps using fastcgi_params, so regardless of the fact that it works or not, the current version is not equivalent to the old version.

docs/websites/nginx/nginx-php-fpm-debian-8.md
```diff
@@ -0,0 +1,165 @@
...
+        location ~ [^/]\.php(/|$) {
+          fastcgi_param REQUEST_METHOD $request_method;
+          fastcgi_pass unix:/var/run/php5-fpm.sock;
+          fastcgi_index index.php;
+          include fastcgi_params;
+        }
+
+Restart nginx by running:
+
+        /etc/init.d/nginx reload
```
FelicianoTech: Let's use `systemctl restart nginx` here.

docs/websites/nginx/nginx-php-fpm-debian-8.md
```diff
@@ -0,0 +1,165 @@
...
+#Packages installation
+In this step, we are installing all the necessary packages.
+Run the following as root to install nginx, PHP and MySQL connector:
+
+        apt-get update
+        apt-get install nginx php5-fpm php5-mysql
+
+Make sure nginx and PHP-FPM are running
+
+        ps aux|grep -e nginx -e php
```
FelicianoTech: We can use `systemctl status nginx` and `systemctl status php-fpm` here. This is more of a discussion point however. If you believe ps aux is the better choice then by all means stick with that.

docs/websites/nginx/nginx-php-fpm-debian-8.md
```diff
@@ -0,0 +1,165 @@
...
+Now we have to setup nginx server to relay on PHP-FPM for PHP requests. 
+By default, PHP-FPM on debian listens on a socket. We are putting everything in one single server so this is ok.
+To find the path for the socket run:
+
+        cat /etc/php5/fpm/pool.d/www.conf |grep 'listen ='
+
+Probably, it is located in `/var/run/php5-fpm.sock`
+
+We will tell nginx where is this socket in a *server* context in the nginx setup.
+In other words, add following lines to `/etc/nginx/sites-enabled/default`, below start of *server* section:
```
FelicianoTech: More clarification on where to put the location block is needed here. Either by mentioning line numbers, or showing a small snippet of what would come before and after where this block needs to be placed.

docs/websites/nginx/nginx-php-fpm-debian-8.md
```diff
@@ -0,0 +1,165 @@
...
+        /etc/init.d/nginx reload
+        
+#Test
+Done! Let's test the web server. Just write the following content in /var/www/html/index.php:
+
+        <?php
+        phpinfo();
+        ?>
+
+And now go to http://example.com/index.php and, if everything is ok, you will see the phpinfo for the installed version.
```
FelicianoTech: This needs to explain that they need to go to their Linode's IP address, the FQDN of the Linode, or another working pointer in order to test this. Some people will actually try example.com while others will replace it with a domain name of their choosing but that doesn't actually point to their Linode yet.

FelicianoTech:
@algspd - As of today, I ran through the guide step-by-step, copying and pasting commands as to not accidentally change anything, and the phpinfo test doesn't work.

Perhaps my location block is placed incorrectly? (Nginx seems to serve it's default page fine).

FelicianoTech:
Hey @algspd, do you have any update for this?

algspd:
I'm too busy right now, I'll check it next month.
You can try to solve it and merge-request the solution.

FelicianoTech:
Sure thing, I'll follow up with you in a month.

By the way, in case you missed it, we have increased the bounty since this pull request was first opened. New submissions can now earn up to $250. The contribute page has been updated to reflect this: https://www.linode.com/docs/contribute

FelicianoTech:
Hi! I just wanted to follow up with you regarding this pull request (PR). Will you have time in the near future to address the questions for the PR?

algspd:
Yep, I'll be able to answer next week (just finishing my exams) :)

docs/websites/nginx/nginx-php-fpm-debian-8.md
algspd: In fact: systemctl status php5-fpm

algspd:
I just fixed your proposals. I'll follow the guide, make additional changes and publish it.

algspd:
There was an error on default site file. Fixed! New version submitted by email, will be online soon.

ellekrout:
Hey @algspd!

Running through another tech test for your guide, the section on logging needs to be altered to remove the suggestion to turn off logging. The "less dramatic" portion of the logging section is perfectly acceptable, but since it's not best practices to turn off logging, and we don't want to mislead newer sysadmins, that portion of the guide needs some editing.

Thank you!

algspd:
Just submitted a fixed version without the logging section.

ellekrout:
Thanks, @algspd! I'll be reviewing the changes later today or tomorrow morning, so we can (finally!) get this pushed to a copy edit, and hopefully published within the next week or two.

Thank you for your patience!

ellekrout:
And hello again, @algspd! I have a PR (thetechstewart/bounty-hiperfnginx/23) with some basic formatting changes, and a couple tech edits. If you'd rather I email it for your review, please let me know! Otherwise, if everything looks good, we'll merge it and then begin the copy editing phase.

Thanks!

algspd:
Just checked the PR, everything ok :) Go on with the merge.
Thanks!

alexfornuto:
Assigning to @kcraig34 for a copy edit.

alexfornuto:
Additional information requested  via email.

algspd:
No email received

cwlinode:
Last correspondence with author was Oct. 28th by email. Considering abandoned. Closing.

PR #232, opened 2015-06-02
Lish commands

alexfornuto:


alexfornuto:
@ellekrout was nice enough to perform a quick copy edit for me.

FelicianoTech:
I like it. Explains the scenarios that may otherwise only be found by accident.

PR #235, opened 2015-06-05
Update yesod-nginx-mysql-on-debian-7-wheezy.md

sheganinans:
Use parallel builds for great good.

### DIFF
--- a/docs/websites/frameworks/yesod-nginx-mysql-on-debian-7-wheezy.md
+++ b/docs/websites/frameworks/yesod-nginx-mysql-on-debian-7-wheezy.md
@@ -94,7 +94,7 @@ To avoid this problem, the maintainers of Yesod created a metapackage named *yes
 
 8.  Now you can install the ``yesod-platform`` and ``yesod-bin`` packages:
 
-        cabal install --reorder-goals --max-backjumps=-1 yesod-platform yesod-bin
+        cabal -j install --reorder-goals --max-backjumps=-1 yesod-platform yesod-bin
 
     The ``yesod-bin`` package provides scaffolding, devel server, and some simple code generation helpers for your ``yesod-platform``. We will use it in the next section to construct the scaffold for a new site.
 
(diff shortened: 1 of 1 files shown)


thetechstewart:
@sheganinans:  I'm sorry for us getting back to you after such a long wait.  Could you confirm what the -j flag that you are passing actually adds to those commands, so that we can correctly review this revision?  I haven't been able to locate any specific information on that flag itself.

sheganinans:
The flag enables parallel builds:
https://ghc.haskell.org/trac/ghc/ticket/910

thetechstewart:
Based on the fact that the only real information I'm able to find on parallel builds directly referencing cabal seems to be on issue reports (admittedly somewhat older reports) I'd like to know what tangible benefits parallel builds would provide to users of this guide before I pull in this request.  Can you provide evidence that utilizing that flag significantly changes build time for Yesod on a Linode?

sheganinans:
Yes, absolutely. It's still a bit new of a feature, hasn't yet gotten the proper docs it needs, and is more of a "tribal knowledge" thing right now. But yes, it absolutely speeds up build times by a whole bunch and is highly recommended. 

Ask these main players in the Haskell community, don't take my word for it:
@bos @tibbe @byorgey @cartazio @chrisdone

thetechstewart:
@sheganinans: I just did a test on a fresh Debian 7 installation, and got the following error when I got to your modified steps.

root@yesodtest:~# cabal -j install --reorder-goals --max-backjumps=-1 yesod-platform yesod-bin
cabal: unrecognized option `-j'

sheganinans:
You need to have a recent version of cabal.

thetechstewart:
@sheganinans - if you'd like to submit updates to the guide in question explaining how to install an updated version of cabal, and including the parallel builds flag, we'd certainly welcome that.  As it stands, since the version being installed by the Debian repositories doesn't support that flag, we would not be able to merge this pull request.

sheganinans:
Ok, sounds good. I think you can use the unstable branch, but since it's not yet in stable it seems like a moot point. Thanks for entertaining the PR, maybe when the right version of cabal is in the stable repo then this will be more relevant.

cartazio:
Parallel package builds via cabal are aok.  Don't use parallel module builds via ghc though.  Im not sure what the larger context of this thread is though.

thetechstewart:
@cartazio:  Thanks for commenting!  I'll give this a test with Debian 8 when I get a chance, and see what needs to be done to get the unstable branch working.  We may plan an update for this guide in the future based upon this commit.

@sheganinans: We definitely appreciate the feedback and the input, the community participation helps keep our guides on the cutting edge.  I'll definitely give this guide a once over with the newer version of Debian, and see if the version of Cabal installed does support parallel builds.

PR #238, opened 2015-06-08
Added StartSSL free ssl guide

sixdev:


sixdev:
And my name is SixDev

alexfornuto:
We already have a guide that covers StartSSL certificates in greater detail: https://www.linode.com/docs/websites/nginx/startssl-wth-latest-nginx-debian-7

Thanks for your submission though!

PR #248, opened 2015-06-30
change '-' to '.' in CentOS6 kernel install cmd

mlewald:
Line 511: When I used  'yum install kernel-x86_64'   the debug version of the kernel was installed.  I changed the hyphen to a dot and the correct non-debug version was installed.  Assuming same for other entries.


alexfornuto:
Thanks @mlewald

PR #250, opened 2015-07-10
Update email-with-postfix-dovecot-and-mysql.md

plinss:
Added an email2email mapping for alias lookups. This allows catch-all email aliases (e.g. @example.com -> email1@example.com) to operate properly. Without this change, a catch-all alias will redirect all email sent to that domain, including email sent to other valid users, because the alias matches before the user lookup. This adds a user lookup to the alias search to email sent to valid users will be delivered to them first.

### DIFF
--- a/docs/email/postfix/email-with-postfix-dovecot-and-mysql.md
+++ b/docs/email/postfix/email-with-postfix-dovecot-and-mysql.md
@@ -287,7 +287,7 @@ Next, set up Postfix so the server can accept incoming messages for the domains.
       query = SELECT 1 FROM virtual_domains WHERE name='%s'
       ~~~
 
-4.  Create the `/etc/postfix/mysql-virtual-domains.cf` file, and enter the following values. Make sure you use the `mailuser`'s password and make any other changes as needed.
+4.  Create the `/etc/postfix/mysql-virtual-mailbox-maps.cf` file, and enter the following values. Make sure you use the `mailuser`'s password and make any other changes as needed.
 
     {: .file }
     /etc/postfix/mysql-virtual-mailbox-maps.cf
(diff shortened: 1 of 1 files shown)


thetechstewart:
@plinss:  Thanks for the contribution!  I'm looking into this right now, but if you have a moment, could you add the syntax for setting up the requisite catch-all alias?  If I manage to locate that before your response I'll pull in your changes and make that addition before publishing.

Thanks again!

plinss:
Added in commit ffe8e4e

docs/email/postfix/email-with-postfix-dovecot-and-mysql.md
```diff
@@ -310,29 +311,41 @@ Next, set up Postfix so the server can accept incoming messages for the domains.
...
+        dbname = mailserver
+        query = SELECT email FROM virtual_users WHERE email='%s'
+        ~~~
+
+7.  Save the changes you've made to the `/etc/postfix/mysql-virtual-email2email.cf` file, and restart Postfix with the following command.
 
         sudo service postfix restart
 
-7.  Enter the following command to ensure that Postfix can find the first domain. Be sure to replace `example.com` with the first virtual domain. The command should return `1` if it is successful.
+8.  Enter the following command to ensure that Postfix can find the first domain. Be sure to replace `example.com` with the first virtual domain. The command should return `1` if it is successful.
```
alexfornuto: I'm retesting this guide using your version. I'm not sure if this stems from your changes or it was a preexisting issue, but when I test other virtual domains in this table, this command still returns `1`. Is 1 supposed to be the row in the table where that domain is listed, or is it just a binary answer?

plinss:
This predates my changes, but the 1 is just a binary answer and seems to work. (My only change to that paragraph was updating the numbering.)

docs/email/postfix/email-with-postfix-dovecot-and-mysql.md
alexfornuto: Nevermind, @thetechstewart knew the answer to this: it is indeed a binary response.

PR #252, opened 2015-07-12
TeamSpeak 3 Guide

sosruko:


thetechstewart:
@sosruko - Thanks for reaching out with this guide.  We've already published a TeamSpeak guide on our Guides and Tutorials site, so we would not be able to accept this submission.  Thank you for your contribution, and we look forward to any guides you may be interested in contributing in the future.

PR #255, opened 2015-07-19
CounterStrike:Global Offensive Server 

thetechstewart:


Joeltrane:
I think that offhand this looks worthwhile. It seems like it could use a lot more context and Linode specific instructions. If the commands listed are able to result in a mostly working CS server than I would say that this could fill a much needed void in our Document Library.

agottschling:
I've setup CS:GO servers before and this is a pretty good tutorial. Just goes about the setup process a little differently than I do.

PR #256, opened 2015-07-19
Fix error in dovecot SSL configuration

jellyjellyrobot:
REF: http://www.dovecot.org/list/dovecot/2014-May/096084.html


alexfornuto:
Thanks again @jellyjellyrobot for the catch!

PR #257, opened 2015-07-20
Fix alias database loop error

jellyjellyrobot:
Alias database loop error caused by repeated domains in BOTH mydestination and virtual_alias/virtual_mailbox_domains

REF: http://serverfault.com/questions/491601/postfix-do-not-list-domain-example-com-in-both-mydestination-and-virtual-alias


alexfornuto:
@ellekrout I believe you worked on this guide last, would you care to look into this?

alexfornuto:
Thanks @jellyjellyrobot !

PR #262, opened 2015-07-28
Deploying a LAMP stack fast and easily with Ansible

rbbrasil:
Hello guys,

Here is my article about Ansible.

### DIFF
--- a/docs/applications/ansible/ansible-lamp.md
+++ b/docs/applications/ansible/ansible-lamp.md
@@ -1,0 +1,294 @@
+# Deploying a LAMP stack fast and easily with Ansible
+
+What's your definition of easy? And what about fast? Well, no matter what answer you have, Ansible can accomplish that! Sounds a little bit arrogant no?, but at the end of this paper I could bet that it accomplishes those goals.
+
+For those who don't know about Ansible, it's one of the most simple and efficient Configuration Manager tool. The short learning curve and the _almost_ non-dependencies on the server-side are the most attractive characteristics of Ansible. Actually, two to four hours of reading the direct and detailed [documentation](http://docs.ansible.com/ansible/index.html) and you're ready to play with your servers!
+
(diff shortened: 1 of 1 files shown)


ellekrout:
Hi @rbbrasil!

Just letting you know that I will be completing a technical test with regards to this article soon, and you should hear back within a week with any comments or changes that should be made.

Thank you for your patience!

rbbrasil:
No problem, @ellekrout 

Thanks for the feedback!

ellekrout:
Hello again!

I have a question before I continue testing the guide: Initially it's presented as being created for Debian 8, but when it comes to deploying the actual playbook, things switch over to Red Hat distros. Is there a reason for this switch? It would be more cohesive if one distro was chosen or the guide included instructions for both.

Just want to get this cleared up before we proceed. Thank you!

rbbrasil:
Actually, it's written for Debian client (line 7) and CentOS/RedHat target host (line 68). If you think that could cause some confusion, I can change the commands/instructions and work only with one type of distro.

I just wrote that way to show that Ansible playbooks has no limitations when running between distributions of different families on the client x server side.

ellekrout:
Ah! Yes, there is some confusion there, but I think if it is defined in the opening paragraphs that we're working with a Debian client and CentOS/RH target host, we'll be good to go.

rbbrasil:
Hi @ellekrout !

Is there any other considerations on this article?

thetechstewart:
@rbbrasil:  Unfortunately, after a great deal of consideration, we were unable to accept this guide for inclusion in the Linode Guides and Tutorials.  We do appreciate your submission, and we encourage you to reach out again if you'd like to submit articles on any other topics.

rbbrasil:
@thetechstewart, could you explain me the reasons? It had been in analysis for more than 20 days and I would be pleased to know why, just in case I decide to submit another one.

PR #272, opened 2015-08-18
Asterisk on CentOS 7

thetechstewart:


capecodrailfan:
It is important to note that when using Asterisk like this SIP/RTP traffic is routed over the internet without encryption. This means calls can be intercepted by anyone between the SIP client and the Asterisk server.

In this case if the Asterisk server is located on a Linode and ISP or monitoring equipment installed by an ISP can intercept the call audio.

To secure this traffic SRTP (Secure Real-time Transport Protocol) can be used.

thetechstewart:
@capecodrailfan:  Thank you for your input!  I'll pass this on to the original author.

PR #273, opened 2015-08-18
Zimbra on Ubuntu 14.04

thetechstewart:


capecodrailfan:
I could be wrong, but that last time I fired up a test Zimbra instance Zimbra is updated by using the same process. A section on updating Zimbra to the latest version might also be helpful.

If a user following this guide intends to use this as a production mail server, a signed SSL certificate should be installed and reverse DNS should be set correctly. If a signed SSL certificate or reverse DNS are not configured properly mail deliver-ability maybe affected.

I have seen in the past where major Internet mail providers will reject mail from servers with improperly configured reverse DNS.

It might also be a good idea to note that DKIM, DMARC, SPF can be used to try and prevent spoofed mail.

BillBardon:
I will add notes on upgrading and installing an SSL cert. The article already mentions the need for a reverse DNS entry, and points readers to the Linode guide on running mail servers, which gives more detail.

Thanks for the comments!

PR #274, opened 2015-08-22
Some VPN guide improvements tested with Debian 8.1

agorf:
Most of them, if not all, should be OK for 7.1 as well.


cwlinode:
Thank you for your input. These improvements were taken into consideration for a new Debian 8 VPN guide which will be published shortly. Closing.

PR #279, opened 2015-09-02
Updated SSL cert and key config line entries

ricknoelle:
The format for the location of the SSL cert and key file entries has changed in Dovecot's config file.  I'm not sure which version the change took place in but I'm running version 2.2.10 on Cent OS 7 and had to use the format specified in my change (or Dovecot would fail to respond to pop3 connections).  This is a great guide, thanks.  This was the only "gotcha" I ran into so I wanted to contribute the update.


ricknoelle:
Here is a reference to the new format:
http://wiki2.dovecot.org/SSL/DovecotConfiguration

alexfornuto:
Thanks @ricknoelle!

PR #280, opened 2015-09-08
update service command to systemctl for CentOS 7

Jirapong:
It seems like this guideline all up to date with CentOS 7. only the fail2bin restart command.

### DIFF
--- a/docs/security/securing-your-server.md
+++ b/docs/security/securing-your-server.md
@@ -347,7 +347,7 @@ Here's how to install and configure Fail2Ban:
 3.  Set the `bantime` variable to specify how long (in seconds) bans should last.
 4.  Set the `maxretry` variable to specify the default number of tries a connection may be attempted before an attacker's IP address is banned.
 5.  Press `Control-x` and then press `y` to save the changes to the Fail2Ban configuration file.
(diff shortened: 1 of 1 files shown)


docs/security/securing-your-server.md
```diff
@@ -347,10 +347,10 @@ Here's how to install and configure Fail2Ban:
 3.  Set the `bantime` variable to specify how long (in seconds) bans should last.
 4.  Set the `maxretry` variable to specify the default number of tries a connection may be attempted before an attacker's IP address is banned.
 5.  Press `Control-x` and then press `y` to save the changes to the Fail2Ban configuration file.
-6.  Restart Fail2Ban by using `sudo service fail2ban restart`.
+6.  Restart Fail2Ban by using `sudo /bin/systemctl restart  fail2ban.service`.
```
alexfornuto: Couldn't this be further simplified as just `systemctl restart fail2ban`? Also, with current SystemD systems like CentOS  does the service also need to be enabled to be reboot persistent?

Jirapong:
Yes. I agree. I will add systemctl enable fail2ban.service as well.

cwlinode:
Added separate lines for systemd and upstart & sysv.
c77c770977957892626ce01e3c8d807844ee7c75

PR #284, opened 2015-09-09
Should be a .crt not .pem

illblew:
I think this was just a typo :question:


thetechstewart:
Good catch @illblew!  Thanks!

PR #287, opened 2015-09-17
Update lamp-server-on-centos-7.md

jlederman:
updated /etc/php/php.ini to /etc/php.ini


ellekrout:
Thanks for the fix, @jlederman! Merging it now. :+1:

PR #289, opened 2015-09-19
OWA Install CentOS v1

dcolby:
This is version 1 of a tutorial for installing Open Web Analytics on CentOS 6.5 and 7.


docs/uptime/analytics/open-web-analytics.md
```diff
@@ -0,0 +1,139 @@
...
+contributor:
+    name: Douglas Colby
+external_resources:
+ - '[OWA Website](http://www.openwebanalytics.com)'
+ - '[OWA Forum](http://www.openwebanalytics.com/?page_id=4)'
+---
+
+##How to Install Open Web Analytics on CentOS 6.5 and 7 (with MySQL back-end)
+
+This guide assume that you have your Linode already set up and running and that you are familiar with the concept of the command line interface (CLI). Since this guide concerns the installation and configuration of software packages all commands are to be run as the root user.
```
alexfornuto: Any guide on installing `X` should start my explaining what `X` is, and why the reader would want to install it.

alexfornuto:
Waiting on author to respond to tech issues via email.

dcolby:
Hi Alex,
I don't see any message in my inbox.

cwlinode:
Hi Douglas,

We're making progress on your OWA guide but there are several issues we've come up against.

* Step 2. chkconfig will not turn on a service immediately for CentOS 7, so any mysql commands won't work because it's not running. Chkconfig, though prominent in previous CentOS versions, is also temporary fallback in CentOS 7 for controlling systemd services. We prefer to encourage readers to use systemd commands when working on distros with it.

* Please add a section to get mariadb (and apache) enabled and switched on.

* I'm getting stuck at the login screen after setting the password for the admin user.
User Name: admin
Password: the pw I made on the previous screen.

Trying to log in then just takes me back to the same screen, though with the text entry fields emptied. Was this something you encountered during your testing?

dcolby:
> On 11 Nov 2015, at 14:53, cwlinode <notifications@github.com> wrote:
> 
> Hi Douglas,
> 
> We're making progress on your OWA guide but there are several issues we've come up against.
> 
> Step 2. chkconfig will not turn on a service immediately for CentOS 7, so any mysql commands won't work because it's not running. Chkconfig, though prominent in previous CentOS versions, is also temporary fallback in CentOS 7 for controlling systemd services. We prefer to encourage readers to use systemd commands when working on distros with it.
> 
> Please add a section to get mariadb (and apache) enabled and switched on.
> 
Will do!
> I'm getting stuck at the login screen after setting the password for the admin user.
> User Name: admin
> Password: the pw I made on the previous screen.
> 
> Trying to log in then just takes me back to the same screen, though with the text entry fields emptied. Was this something you encountered during your testing?—
> 
I did not encounter this problem but it was some time ago when I tested it. I will investigate—maybe a CentOS update has introduced some snags.

DC

dcolby:
Added instructions for starting MariaDB and Apache automatically on boot. Added note that OWA requires a fully qualified domain name and not an IP address (the source of the login page round-a-bout).

docs/uptime/analytics/open-web-analytics.md
```diff
@@ -0,0 +1,150 @@
...
+    mysql -u root -p
+
+4. The following SQL commands will create a database named `owadb` with a user named `owadbuser` and with the password `owadbpassword`. Be sure to change the password to something better of your own choice. This information will be needed later to configure OWA.
+
+    CREATE DATABASE owadb;
+    GRANT ALL PRIVILEGES ON owadb.* TO owadbuser@localhost IDENTIFIED BY 'owadbpassword';
+    FLUSH PRIVILEGES;
+    quit
+
+5. Now you need to restart your Linode:
```
alexfornuto: Why would you need to reboot your Linode at this point? If it's to start the services, that won't matter because:
a) you need to start the mariadb service before you can do any of this section, and 
b) you can start both services from the command line.

dcolby:
I agree that it is not really necessary. I just like to check that everything runs automatically after rebooting.

dcolby:
Maybe it should say: I like to restart my Linode at this point to check that the services run at startup but it is not necessary.

alexfornuto:
I'm going through this guide now, and expanding it to be a little more distro-agnostic. Can you explain in a sentence or two why OWA has a limitation on access via the IP address?

docs/uptime/analytics/open-web-analytics.md
```diff
@@ -0,0 +1,150 @@
...
+
+	Database User: owadbuser
+
+	Database Password: owadbpassword
+
+Of course, `owadbpassword` should be the password with which you created your SQL database in step 2.4.
+
+![Screen shot of OWA set up screen.](/docs/assets/OWA.png)
+
+3. Click "Continue..." and create a user account on the next page. This user account will be how you log in to see your OWA statistics.
```
alexfornuto: During my tech testing it appears that this step in the installation page also has us create the first site to track, so that when we log in we already have some code presented. This affects both this step and the next section.

alexfornuto:
Hello @dcolby,

In addition to my notes above, I've made a PR for you with changes: https://github.com/dcolby/docs/pull/1

Please update us once you've merged my PR (or if you have any questions about it first) and addressed the outstanding notes.

Thanks!

docs/uptime/analytics/open-web-analytics.md
```diff
@@ -0,0 +1,188 @@
...
+
+
+
+2.  After clicking on **Let's Get Started**, you should see a configuration page for your OWA installation. The first field will be filled in by OWA with the path to your OWA installation. The other fields on the page should be filled in with the information you set in the MySQL CLI:
+ 
+    [![Open Web Analytics set up screen.](/docs/assets/owa-install_small.png)](/docs/assets/owa-install.png)
+
+3.  Click **Continue...**. 
+
+4.  Create a user account and define a domain to track. This user account will be how you log in to see your OWA statistics.
```
alexfornuto: I made this note on what is now an outdated diff, but it still applies:

During my tech testing it appears that this step in the installation page also has us create the first site to track, so that when we log in we already have some code presented. This affects both this step and the next section.

alexfornuto:
Hey @dcolby the line note I just added covers just about the last phase of my testing. Once we take care of the discrepancy between the last section and the practical test, we can move this into copy edit, and get it published!

alexfornuto:
Thanks @dcolby for the changes! I've asked @kcraig34 to perform a copy edit and review, and then you should be good to go! 

In preparation for publishing, would you prefer payment or a credit to your Linode account? In either case, we'll need the associated email address to use.

alexfornuto:
Thanks for a great contribution @dcolby! Your guide will be published today, and you should receive payment within 24 hours.

PR #299, opened 2015-09-28
fist -> first

CraigyDavi:


ellekrout:
Oh man! Thanks for catching that for us! We'll merge it right away! :)

PR #275, opened 2015-08-24
StartSSL with Apache on Debian or Ubuntu

thetechstewart:
This is an import from a submission given to us via email. It has not yet been reviewed for technical accuracy or edited for for proper formatting. It's purpose here is to gauge interest from the community on this subject, and solicit independent review of the contents.

PR #304, opened 2015-10-02
Configuring StartSSL with Apache on Debian or Ubuntu

dreamerblue:
This is a guide about StartSSL and its application. The original version refers to PR [#275](https://github.com/linode/docs/pull/275).


alexfornuto:
What are the pros / cons of using the StartSSL "Express Lane" versus signing up for an account?

alexfornuto:
Unfortunately I'm unable to review this because I already have a StartSSL account. The next available team member will take a look.

dreamerblue:
@alexfornuto @cwlinode Thanks for suggestion. I have modified the description of Express Lane.

cwlinode:
@dreamerblue 

We've decided to reorganize and restructure our guides which deal with TLS certificate installations. Since Let's Encrypt is publicly available now, we're also choosing to focus on that project for exemplary purposes rather than StartSSL Free.

We'll preserve elements from our current guide on nginx with StartSSL for readers still interested in that certificate brand, but unfortunately we will not be adding any further material on StartSSL at this time.

PR #310, opened 2015-10-07
Corrected word spellings

tylerjaynelson:


ellekrout:
Thanks for the catches! Fixing. :)

PR #313, opened 2015-10-09
Update server download link

Ybrin:
Changed server download link to point to version 1.8.8 as this is more stable.


alexfornuto:
Thanks @Ybrin!

PR #321, opened 2015-10-27
Update limiting-access-with-sftp-jails-on-debian-and-ubuntu.md

benny-gold:
There's a mis-capitialisation in the /etc/ssh/sshd_config file that could potentially lock a user out of their node


felicianotech:
According to OpenSSH's [man page](http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man5/ssh_config.5?query=ssh_config&sec=5):
>The possible keywords and their meanings are as follows (note that keywords are case-insensitive and arguments are case-sensitive):

With that being said, it does look like the convention is to use a capital for the first letter of keywords. Thank you for your fix. Much appreciated.

PR #332, opened 2015-11-05
Should be "GATEWAY" instead of "GATEWAY0". Network configuration fail…

fringerdarlinger:
…s with "GATEWAY0".


alexfornuto:
Thanks @fringerdarlinger! We're rolling this into a bunch of other changes to the guide that should be published soon.

cwlinode:
fixed/updated: https://www.linode.com/docs/networking/linux-static-ip-configuration

PR #335, opened 2015-11-10
Fixing an error and providing explanatory text

UltraBob:
As it currently stands the instructions tell you to create certificates and then chmod some unrelated files that probably don't exist.  I changed the chmod command to relate to the files generates, and added some text to explain that example.com should be replaced with the domain in question in both commands.


alexfornuto:
Good catch, thanks @UltraBob!

PR #339, opened 2015-11-12
default IPv6 address not assigned by SLAAC when /64 pool requested

avolkov:
I verified that after requesting /64 pool, the default address is no longer assigned by SLAAC in Debian Jessie. The administrator will need to statically define that address in /etc/network/interface, failing to do that will break IPv6 routing.

### DIFF
--- a/docs/networking/native-ipv6-networking.md
+++ b/docs/networking/native-ipv6-networking.md
@@ -106,6 +106,9 @@ While default IPv6 addresses are configured automatically, you will need to stat
         address 2001:DB8:2000:aff0::2/64
       ~~~
 
+{: .note}
+>On Debian Jessie, your default IPv6 address provided by SLAAC will no longer be automatically assigned after you request /64 pool. You will need to manually add it as a static address or IPv6 routing will not work.
+
 2.  Restart networking. As this will break an SSH connection, this command should be performed in [LISH](/docs/networking/using-the-linode-shell-lish)
 
         ifdown -a && ifup -a


alexfornuto:
Good catch! Before merging this I would like to see if this affects all IPv6 pool users, or just those running Jessie.

Jnosh:
Maybe that's just me but I also had a problem with SLAAC not working on Jessie after following the guides.

Adding the firewall rules mentioned in 'docs/security/firewalls/control-network-traffic-with-iptables' ultimately solved the problem for me.

I created a pull request to add these rules to the 'docs/security/securing-your-server' guide but I'm not sure if all systems need these rules added separately or just Debian 8?

tknarr:
From the looks of the changes and my fun with IPv6 those rules should be appropriate for any Linux system, the only exceptions I can think of being ones that have an IPv6-enabled firewall running already.

[RFC 4890](https://www.ietf.org/rfc/rfc4890.txt) and [RFC 6092](https://www.ietf.org/rfc/rfc6092.txt) have recommendations, especially 4890 section 4.4 where it discusses what ICMPv6 packets have to be allowed through for endpoint equipment.

alexfornuto:
Hey @avolkov and others. I'm working on updating this guide for all distributions. My current commit (https://github.com/alexfornuto/docs/commit/d1d6169004105efd89261b693d86405c4aafdf20) works for Debian, but not Ubuntu. Ubuntu prefers when we add addresses as shown in the current version, and will also provide the SLAAC address along with the statically assigned numbers from the pool.

So this is where I'm left pondering, and I would like to open it up for discussion; should we address each IPv6 address with an `iface eth0 inet6 static` block and tell users that they need to statically assign the default address as well, or split the Debian & Ubuntu section for each distro, with Debian using `up      ip -6 addr` lines?

alexfornuto:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;qFjnsdHk&#x2F;6-ipv6-address-assignment"> IPv6 Address Assignment</a>

PR #340, opened 2015-11-12
Streaming Audio

falvert:
First guide for streaming audio with socket.io.

### DIFF
--- a/docs/applications/media-servers/sioaudio.md
+++ b/docs/applications/media-servers/sioaudio.md
@@ -1,0 +1,258 @@
+---
+author:
+    name: Linode Community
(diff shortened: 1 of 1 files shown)


cwlinode:
@falvert Thank you for your submission.

After some internal discussion, we have chosen not accept this guide. While we would definitely like a tutorial for creating an application using Socket.io, we would prefer a more extensive use case than streaming a media file.

However, if you have any other topics in mind, please review our Writer's Guides and submit a new PR.
https://linode.com/docs/linode-writers-guide
https://linode.com/docs/linode-writers-formatting-guide

PR #341, opened 2015-11-17
Corrected command on configuring firewall with ufw that was improperl…

blhylton:
…y listed as a service instead of an argument to the binary


cwlinode:
@blhylton Thank you. Edits added for clarity.

PR #346, opened 2015-11-26
Installing Kurento Media Server with TURN server on Ubuntu 14.04

spariev:


cwlinode:
@spariev Thank you for your submission, but we have chosen not to accept this guide. We feel that between Kurento's official documentation and other sites on the internet, the included subjects are sufficiently covered.

spariev:
@cwlinode thanks for letting me know. Just to clarify, I wrote this tutorial because `coturn` TURN server, which is suggested in Kurento official docs, is very tricky to install on Ubuntu as it only has Debian packages, so I thought showing how to use Kurento with reTurn would be of some use.

PR #347, opened 2015-11-28
Fix switched instructions

john-pettigrew:
It seems like these instructions were switched.


dbinoj:
was about to fix it. luckily found your PR.

cwlinode:
Thanks for catching this, guys. Fixed.

PR #349, opened 2015-11-30
Installing Java on Ubuntu 14.04

seanthewebber:
Completed tutorial; creating pull request momentarily. This will act as
a prerequisite to a tutorial I want to write later this week.


cwlinode:
@seanthewebber Thank you for your submission, but we have chosen not to accept this guide. We aim to avoid duplicating information whenever possible and there is already plenty of instruction on the internet for Java installations in Linux and setting the Home environment.

seanthewebber:
@cwlinode I understand. However, this guide served as a prerequisite to #353. Should I condense this guide and add it as a new section to that one?

PR #350, opened 2015-11-30
Fix syntax error in 'create user' command

nikolas:
Fixed the error: "You have an error in your SQL syntax;" in the example create user command.


cwlinode:
@nikolas Thank you for catching this.

PR #351, opened 2015-12-01
Update custom-distro-on-kvm-linode.md

dadooda:
Remove period after end of title.


cwlinode:
@dadooda Thank you for catching this.

PR #356, opened 2015-12-11
Add iptables rules for IPv6 SLAAC

Jnosh:
The rules were lifted verbatim from 'docs/security/firewalls/control-network-traffic-with-iptables'.

I'm no iptables expert but at least for me these rules were necessary to make SLAAC work on Debian 8.
It would be great if someone more knowledgable in this area could chime in whether this makes sense for all systems?

### DIFF
--- a/docs/security/securing-your-server.md
+++ b/docs/security/securing-your-server.md
@@ -354,6 +354,12 @@ If you would like to supplement your web server's IPv4 rules with IPv6 too, this
     -A INPUT -i lo -j ACCEPT
     -A INPUT ! -i lo -s ::1/128 -j REJECT
 
+    # Below are the rules which are required for your IPv6 address to be properly allocated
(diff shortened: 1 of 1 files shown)


alexfornuto:
Thanks for the suggestion, and sorry for the late reply! We did thorough testing, and spoke with our network admins, and we finally reached a conclusion.

These specific rules re-allow for SLAAC addressing, assuming that you've otherwise blocked ICMPv6 traffic, and restrict the number of hops to prevent spoofing. Our network infrastructure already prevents spoofing, and the ip6tables rules that we list in the guide allow all ICMPv6 traffic, negating the need for these more specific rules. We will most likely remove these rules from the iptables guide, since they're extraneous. 

Several members of my team, myself included, tested the rule set from our Securing Your Server guide and found that we still are able to get SLAAC addresses without these rules. I tested Debian 8 personally. 

Thanks for your time! Please let us know if you are still encountering any issues.

PR #357, opened 2015-12-11
riak cluster on ubuntu 14.04 guide

spariev:


cwlinode:
@spariev Thank you for your submission, but we have chosen not to accept this guide. We feel that the the official documentation on docs.basho.com is sufficient, and we do not want to host redundant information when such in-depth instruction already exists.

PR #359, opened 2015-12-16
fix typo, iptop -> iotop

titusnicolae:
change iptop to iotop


cwlinode:
@titusnicolae Thank you for catching this.

PR #362, opened 2015-12-19
Installing with Brew commands need `install`

ntwb:


felicianotech:
Great catch! Thank you.

PR #358, opened 2015-12-12
Installing Roundcube on Ubuntu 14.04

seanthewebber:
- Added metadata heading
- Added numbering to comply with the Linode Style Guide
- Fixed unmet PHP PEAR requirements discovered in pull request #223 (since closed)
- Improved phrasing and section descriptions
- Updated Roundcube version from 1.1.1 to 1.1.3

PR #366, opened 2015-12-22
Obtaining Let's Encrypt Certificates on Ubuntu 14.04

seanthewebber:
Finalized guide submission for obtaining Let's Encrypt SSL certificates on Ubuntu 14.04.


docs/security/ssl/obtaining-lets-encrypt-certificates-on-ubuntu-14-04.md
```diff
@@ -0,0 +1,190 @@
...
+- Required attention and maintenance
+- Technical details about Let's Encrypt and certificates issued by Let's Encrypt
+
+{: .caution}
+>
+> As of 2016-01-01, Let's Encrypt is still in *public beta*. Although most users have reported success with Ubuntu 14.04, the ACME client is still being debugged and developed. **Do not deploy Let's Encrypt Public Beta in a production environment without testing it beforehand.**
+
+## Prerequisites
+
+- A Linode server running Ubuntu 14.04 with < 2GB of RAM. On a **Linode 1024** server, add 1GB of SWAP to meet this requirement. We recommend following our [Getting started](/docs/getting-started) guide if you need help setting up your Linode
```
zupolgec: with "less than" 2GB of RAM?
apart from that, why this requirement?

seanthewebber: It's due to a limitation I discovered when I tried testing the guide on a Linode 1024 server. `letsencrypt-auto` compiles some libraries when it initializes a virtual environment, and cryptography` fails to compile when less than 2GB of total memory is present. It proceeds to throw an `InsecurePlatformWarning` because of the uncompiled library.

Source: [InsecurePlatformWarning · Issue #1883 · letsencrypt/letsencrypt](https://github.com/letsencrypt/letsencrypt/issues/1883)

zupolgec: @seanthewebber :+1: 
so just fix the "less than" symbol. should be `>` not `<`, right?

seanthewebber: Instead of this:

`A Linode server running Ubuntu 14.04 with < 2GB of RAM.`

[You mean like this?](https://en.wikipedia.org/wiki/Greater-than_sign)

`A Linode server running Ubuntu 14.04 with 2GB > of RAM.`

zupolgec: I think that the correct way is
`A Linode server running Ubuntu 14.04 with > (more than) 2GB of RAM.`
because the greater amount is the one you have in the VM, the one that should be higher than 2GB.

I would probably rephrase it as
`A Linode server running Ubuntu 14.04 with at least 2GB of RAM.`

seanthewebber: True, but I like the idea of eliminating possible confusion all together. Making the changes now.

seanthewebber:
@cwlinode Any status updates for this guide thus far?

zupolgec:
I started using let's encrypt on a Linode server, but I'm using letsencrypt.sh instead of the official client. Much much simpler and straightforward to use.

seanthewebber:
@zupolgec Can you provide some examples of how to use that? Additionally, is `letsencrypt.sh` an official script written by Let's Encrypt?

zupolgec:
@seanthewebber here is the repo https://github.com/lukas2511/letsencrypt.sh
It's just a script for the letsencrypt ACME server

seanthewebber:
@zupolgec That looks useful! I will keep a note of that for my personal servers.

However, is it a third-party script not maintained by Let's Encrypt. Instructing Linode's readers, most of which run production environments, to rely upon an unofficial script seems like a bad idea. Unless @cwlinode disagrees with that logic?

cwlinode:
@seanthewebber Your guide is near the front of the queue so we should be moving on to it shortly. We'll be in touch with you at that time.

Re the script: We do prefer to stick to official sources whenever possible. At the end of the day though, it comes down to a reader using whatever their choice may be.

cwlinode:
@seanthewebber Thanks for your patience with this. I tech tested this on Debian 8, Ubuntu 14.04 and CentOS7, our core distros, just to be sure there were no unseen hangups. Two things so we can move on to the copy edit phase:

1.  Please remove the Ubuntu reference and add a quick `yum update` to the lines after apt-get so this can be a more universal guide.
2.  The "Exploring..." section needs sudo on the commands because of file permissions.

What are your thoughts on adding a cron job to git pull for updating LE?

seanthewebber:
@cwlinode All requested changes have been completed and committed (see below). Please let me know if further modifications are necessary.

> What are your thoughts on adding a cron job to git pull for updating LE?

I went ahead and added a cron job that executes the `cd /opt/letsencrypt && git pull` command. Once Let's Encrypt leaves public beta, an `apt-get` repository will *probably* be maintained and this command can be removed.

cwlinode:
@seanthewebber Looks good. Has been passed on for copy editing.

seanthewebber:
@cwlinode @kcraig34 Just accepted your merge request.

cwlinode:
@seanthewebber Thank you for your contribution. Your guide is published:
https://linode.com/docs/security/ssl/install-lets-encrypt-to-create-ssl-certificates

Please confirm your Paypal address to contribute@linode.com.

seanthewebber:
@cwlinode Thank you for publishing this! Please direct payment to swebber@yazzielabs.com. Would like to avoid the email hiccup we had in #358.

cwlinode:
@seanthewebber No problem. Payment sent.

PR #367, opened 2015-12-22
Fix typo, rephrase a bit.

demetris:


cwlinode:
@demetris Thank you for catching this.

PR #370, opened 2015-12-28
Hadoop on Ubuntu 14.04

cwlinode:
Submitted by email. Should the guide be accepted, it would be for Trusty instead of Precise.


cwlinode:
Guide rejected. Closing.

PR #371, opened 2015-12-29
Fix typo

justincwatt:
Pretty sure the docs meant to say "prevent" instead of "percent of"


cwlinode:
@justincwatt Thank you for catching this.

PR #372, opened 2016-01-01
Corrected typo in CENTOS 7 static networking config example

dtwist:
Small typo, but makes a big difference. Without the correct suffix here, network no-workie.


cwlinode:
@dtwist Thank you for catching this.

PR #374, opened 2016-01-04
Update set-up-a-hardened-openvpn-server.md

wabzqem:
Fix link to openvpn-access-server


cwlinode:
@wabzqem Thank you for catching this.

PR #382, opened 2016-01-20
fix the error command

krizex:
The *-m* option in command *iptables* has clerical error.


alexfornuto:
Good catch, thank you @krizex!

PR #383, opened 2016-01-22
Fix iptables rules

jclopes:
The logging rules were logging 3x the INPUT and no logging of the FORWARD and OUTPUT.
Added commented out rules to enable a local http/s server


cwlinode:
@jclopes Thanks for catching the INPUT 3x part. As for adding web server rules, that's outsuide the scope of this guide. If someone wants to add their own rules, they're always free to do so at any point.

PR #385, opened 2016-01-27
Update discussion-forums-with-smf.md

jdarwood007:
Always point to the latest installer file.


alexfornuto:
Thanks @jdarwood007!

PR #386, opened 2016-01-27
Install and Running Hadoop on Ubuntu 14.04 (Single-Node-Cluster).

mani90:
Install and Running Hadoop on Ubuntu 14.04 (Single-Node-Cluster).

### DIFF
--- a/docs/applications/hadoop/hadooponubuntu1404.md
+++ b/docs/applications/hadoop/hadooponubuntu1404.md
@@ -1,17 +1,19 @@
 ---
 author:
-	name: Linode Community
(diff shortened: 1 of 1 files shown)


thetechstewart:
@mani90
Upon review of your submission, portions were found to be copied without attribution from an already published source.  Additionally, it appears that this is your second attempt to submit a previously rejected guide.  We will not review any further submissions from you on this topic.

Original Source: http://www.michael-noll.com/tutorials/running-hadoop-on-ubuntu-linux-single-node-cluster/

PR #387, opened 2016-01-27
Fix apparent mistake in IPv6 firewall rules

wturrell:
It appears only accepting ICMP traffic matching `-m state --state NEW` stops all incoming connections – I could not even ping the machine with `ping6`. My knowledge of this area is quite weak, but the suggestion on IRC was that Conntrack does not track state, so --state NEW would not match.

This was using Debian 8.2 with the newest Kernel (4.4.0-x86_64-linode63)


cwlinode:
@wturrell Thank you for this.

PR #392, opened 2016-02-03
Spfdkim rotate testkey

tknarr:
Expand step 2 of rotating keys to include verifying the new key and DNS data before activating the key in Postfix.

Fix a minor naming/numbering error referring to a previous section and step.


alexfornuto:
Sorry it took so long! Everything looks and works well.

PR #394, opened 2016-02-09
Fixed typo in command example

module17:


cwlinode:
@module17 Thank you for catching this.

PR #398, opened 2016-02-25
[SteamCMD] Changed installation folder to the standard one

davidebeatrici:
SteamCMD automatically uses the "Steam" folder for its files (logs, workshop items, etc.). Using another folder to extract files in results in duplicate files and unneeded occupied space.
It's better to keep only one folder, for space and order reasons.
I also changed the user home folder in the installation log to match the suggested user name.


cwlinode:
@davidebeatrici Thanks for this. I'll fix the other Steam guides to reflect the new directory.

PR #407, opened 2016-03-02
Update and rename how-to-install-nginx-and-php-via-fastcgi-on-ubuntu-…

kcraig34:
…12-04-lts-precise-pangolin.md to Install-Nginx-&-PHP-via-FastCGI-on-Ubuntu-12-04.md

Title shortened. URL and meta description aligned. Keywords added. Copy edited.


alexfornuto:
Don't title case the file names. Also, this was taken care of in the mass update PR

PR #415, opened 2016-03-14
Openfire

pbzona:
Did some general cleanup - killed VPS, rewording, and updated intro to specify use as a non-root user.


alexfornuto:
You moved this from a root to sudo use guide. Did you test it?

pbzona:
I didn't test, just updated the necessary commands to use sudo. My thinking was we shouldn't be encouraging people to install software as root unless absolutely necessary, and in this case it isn't.

alexfornuto:
I agree with your line of thinking, but if you're going to make changes to the commands themselves you need to test them.

pbzona:
I'll go ahead and change back to root use then to save time. Do we have a general rule that guides should be written for non-root users unless absolutely necessary? If so, maybe we should lay that out in the writer's guide to promote greater security, even when installing trusted software.

alexfornuto:
Yes, that is a general rule, and would be a good addition to the writer's guide.

alexfornuto:
:+1:

PR #417, opened 2016-03-21
LISH: Remove keysview command

dwfreed:
While keysview is still a valid command, it is no longer relevant with the LISH gateways, so remove it from the guide.


alexfornuto:
@displague can you confirm?

displague:
👍 This is a legacy command that should be axed higher up the chain.

alexfornuto:
Thanks @dwfreed

PR #418, opened 2016-03-21
Add line to fix the appearance of Grub in Lish

decomposite:
Grub doesn't look very good on Lish without the GRUB_TERMINAL=serial line as Lish acts as a serial terminal. Line 133 is just a newline trimmed at the end of the file.


alexfornuto:
Thanks @decomposite

PR #420, opened 2016-03-24
Update configure-firewall-with-ufw.md 

partounian:
Sudo ufw delete allow 22 changed to HTTP 80 as stated in description above.


alexfornuto:
Good catch, thanks @partounian!

PR #421, opened 2016-03-28
Finished guide on how to setup a website on an Intranet

Agent-Isai:
Adding my intranet guide, this guide is for use with Debian-based systems such as Debian itself and Ubuntu.


alexfornuto:
@Agent-Isai Thanks for the submission! Upon review, I don't think this guide is in-depth enough to be considered. Setting a local hosts file to point your loopback address to a dummy domain does not actually create an Intranet, it only lets you see your local apache services using that domain, only on your local computer.

 A better solution would involve having DNS service on the local network define the dummy domain and an IP address for it.

PR #438, opened 2016-04-11
Make adjustments to our Arch Linux networking documentation in accord…

decomposite:
…ance to our current image.

I've added a configuration snippet for Arch Linux IP configuration which does not rely on delegation to Arch's docs. Please note that it uses systemd-networkd instead of netctl as that is what network helper edits.

And I've switched netctl and systemd-networkd, for the Native IPv6 guide, in order to encourage systemd-networkd to be tried first.

This should help avoid any confusion or mistakes, as Network Helper automatically generates good examples to be worked from which will end up being more simple and headache free.


alexfornuto:
Thanks @decomposite!

PR #443, opened 2016-04-17
Use mysqli_real_escape_string

a3020:
The function mysql_real_escape_string was deprecated in PHP 5.5.0, and it was removed in PHP 7.0.0.


alexfornuto:
Nice addition, thanks @a3020

PR #444, opened 2016-04-17
Add guide for running an Omni Sync Server

huw:
OmniFocus and other Omni Sync applications are really useful—so I thought I'd make a guide for hosting a private sync server for them. This could be useful for people who don't want to host their data with OmniFocus and would rather use their Linode.

It's also a decent guide for running a WebDAV server if anyone needs that.


alexfornuto:
@huw  thanks for the submission! Unfortunately we're unable to accept this guide. We've paused the bounty system while we [upgrade](https://github.com/linode/docs#update-072216) the process. The new bounty proposal system will make it much easier to discuss and approve topics before they're written, saving you time and effort in your next submission. Thanks again!

PR #449, opened 2016-04-22
Fix typo 'permanently'

briansteffens:


alexfornuto:
Nice catch, thanks @briansteffens!

PR #453, opened 2016-04-25
added IPv6 addresses for slave servers; removed obsolete addresses fr…

ravenx99:
…om slave servers


alexfornuto:
Thanks @ravenx99!

PR #454, opened 2016-04-27
fix spelling mistake

jschollenberger:
gudielines -> guidelines


alexfornuto:
Good catch, thanks @jschollenberger!

PR #455, opened 2016-04-27
This link is broken

Swashy:
This link is broken, looks like the title was changed at some point

PR #457, opened 2016-05-03
Broken link Fix

alexfornuto:
The original fix from https://github.com/linode/docs/pull/455, with clarified copy.


LiveMNCcom:
Nice Webhosting services

PR #461, opened 2016-05-04
Update configure-spf-and-dkim-in-postfix-on-debian-8.md

p4squ4l3:
Postfix ≥ 2.6 milter_protocol = 6, Postfix ≤ 2.5 milter_protocol = 2


alexfornuto:
@tknarr or @p4squ4l3, would either of you care to elaborate on this change?

tknarr:
The version numbers noted are the default versions (Postfix 2.6 changed the default from protocol version 2 to 6), and either protocol version works correctly so the change should be fine.

p4squ4l3:
Yes, I read and tested
http://www.postfix.org/MILTER_README.html
http://www.postfix.org/postconf.5.html#milter_protocol

alexfornuto:
Thanks @p4squ4l3, and @tknarr for reviewing.

PR #462, opened 2016-05-04
Update configure-spf-and-dkim-in-postfix-on-debian-8.md

p4squ4l3:
added opendkim-genkey algorithm and domain specification


alexfornuto:
@tknarr Would you care to weigh in on this?

tknarr:
I don't see any problem with the changes. `rsa-sha256` is the default alrogithm, making it explicit doesn't hurt, and the domain information from the Bind TXT record output isn't going to be used so it doesn't matter what value it has.

p4squ4l3:
I added domain information only for convenience, the algorithm to avoid untrusted default configuration.

PR #473, opened 2016-05-25
Update getting-started.md

smoreface:
Pretty minor copyedits.

FWIW, I wanted to eradicate the gerund (installing, booting, connecting vs install, boot, connect) for clarity's sake, but I noticed you link throughout other docs to the subheads here, so I didn't want to introduce a lot of maintenance upkeep for a semi-minor clarity quibble.

The Logging In for the First Time section, for me, step 4 happened before step 3. I didn't switch the steps since I wasn't sure if it was just me or not.


pbzona:
I'll go through the Logging In section and see if there's a problem with the instructions. Even if it was just an error in your case, that could suggest a lack of clarity on our part. Thanks for your help!

PR #475, opened 2016-05-26
Fix Systemd Config Highlighting And Spacing

amayer5125:
Make all systemd configs consistant
Set highlighting to ini since it is the closest language
that has highlighting
Remove spaces between = sign and description


EdwardAngert:
Thanks for pointing out the ini highlighting, @amayer5125!
I was able to incorporate your suggestion, but had to do it in another branch to avoid a merge conflict, so while I'm closing this PR, you'll notice that your contribution is still recognized on the merged PR, https://github.com/linode/docs/pull/548.

Thanks again!
Edward

amayer5125:
Thanks! Glad I could help.

PR #479, opened 2016-05-30
Remove ApacheBuddy, which doesn't work

reppep:
[root@linode2 conf.modules.d]# perl ~pepper/apachebuddy.pl 
########################################################################
# Apache Buddy v 0.3 ###################################################
########################################################################
Gathering information...
We are checking the service running on port 80
The process listening on port 80 is /usr/sbin/httpd
The process running on port 80 is not Apache. 
 Falling back to process list...
Could not find Apache process. Exiting...


alexfornuto:
Thanks for bringing this to our attention! It looks like the script referenced has been replaced with Apache2Buddy. I've staged https://github.com/linode/docs/pull/525 which updates this guide to point to the correct script.

PR #267, opened 2015-08-03
Nginx SSL + TLS Deployment Best Practices 

capecodrailfan:
A new guide that builds on the existing guide on deploying the latest version of Nginx with a free SSL certificate from StartSSL.

PR #481, opened 2016-06-01
Nginx ssl tls

cwlinode:


alexfornuto:
This is a continuation of https://github.com/linode/docs/pull/267 since we haven't heard back from the author. Whoever merges this, be sure to resolve the original PR as well (including bounty).

docs/websites/nginx/how-to-install-nginx-and-a-startssl-certificate-on-debian-8-jessie.md
```diff
@@ -1,250 +0,0 @@
----
```
alexfornuto: I am trying to figure out why this deleted file shows up in this PR. I understand that it's an older version of another guide the original author had accepted under a new name. Since that means there's no guide in master by that name, why is it showing as a deletion at all?

In any case, it shouldn't be a problem, but whoever merges this needs to be sure to check this guides' final form, `install-nginx-and-a-startssl-certificate-on-debian-8-jessie.md`

alexfornuto:
Closing this since the original PR is active again.

PR #487, opened 2016-06-04
Command should point to stable release

navjotjsingh:
Shouldn't the guide be about upgrading to a stable release?

### DIFF
--- a/docs/security/upgrading/upgrade-to-ubuntu-16-04.md
+++ b/docs/security/upgrading/upgrade-to-ubuntu-16-04.md
@@ -98,7 +98,7 @@ Remember to perform these steps in a Lish or Glish session:
 
 3.  You're now ready to begin the upgrade to Ubuntu 16.04 LTS:
 
-        do-release-upgrade -d
+        do-release-upgrade
 
     Follow the on-screen instructions to complete the installation process.
 


docs/security/upgrading/upgrade-to-ubuntu-16-04.md
```diff
@@ -98,7 +98,7 @@ Remember to perform these steps in a Lish or Glish session:
 
 3.  You're now ready to begin the upgrade to Ubuntu 16.04 LTS:
 
-        do-release-upgrade -d
+        do-release-upgrade
```
EdwardAngert: Hi @navjotjsingh,
I tried running the upgrade without the -d flag like you're suggesting, but I got an error back.
Was there something else you tried here that worked better than what we've got?


Thanks,
Edward

EdwardAngert: @navjotjsingh ,
Ran through it again to confirm and check on that "No new release found" error (screenshot attached).
After some further research, I found that you're right, but not yet.
The -d flag calls for a development release because Ubuntu 14.04 doesn't recognize 16.04 as an upgrade path until 16.04.1 is released (at some point in July). I'm going to leave the guide as written until 16.04.1 is released, then we'll update to remove the flag.

1__ssh_and_how_to_upgrade_to_ubuntu_16_04_lts

Thanks,
Edward

PR #489, opened 2016-06-08
removed an rm -rf command

Swashy:
changed "cp -R" to "mv" and removed the force remove command. That's dangerous!


alexfornuto:
Sounds like a better choice of commands! Before we can merge, however, the instructions above need to be amended to match the new commands. You can pretty much remove everything after the comma, and change it to a period.

docs/websites/cms/how-to-install-and-configure-wordpress.md
```diff
@@ -78,8 +78,7 @@ WordPress is a popular, dynamic, blog-focused content management system. The sof
 
 5.  Copy the WordPress files to your `public_html` folder, then remove the folder in the `src` directory:
```
dopeghoti: Suggest changing this to match the PR, with:

> 5.  Move the WordPress files to your `public_html` directory:

PR #491, opened 2016-06-09
How to Use Webalizer to Analyze Web Traffic on Debian 8 (Jessie)

Fred-Zweig:
This tutorial explains how to configure Webalizer to process Nginx logs on Debian 8. Webalizer is a popular web traffic analysis tool that has the advantage of being lightweight yet powerful, so that it's capable to deliver in-depth analysis without using any significant amount of resources.


Fred-Zweig:
Although there is [an old Linode guide](https://www.linode.com/docs/uptime/analytics/webalizer-on-debian-5-lenny/) on how to install Webalizer on Debian 5, my guide is not an update to the old one. That guide describes how to install Webalizer to analyze Apache logs on Debian 5, while my guide explains how to install Webalizer to analyze Nginx logs on Debian 8. Unlike the old guide, my tutorial explains in detail each configuration option that must be turned on in the `/etc/webalizer/webalizer.conf` file. It describes how to enable the geolocation service, how to update the list of search engine query strings, how to configure http authentication to restrict access to Webalizer reports, how to properly configure 'logrotate', and the best way to set up the cron job. It describes everything that a user needs to do, to have a really functional Webalizer installation for Nginx and Debian 8. It doesn't send the reader to other guides. For example, I didn't send the user to the Linode guide for 'logrotate' which just describes the general case of using 'logrotate' and it doesn't show how to configure it for Webalizer to really work, with the proper 'prerotate' and 'postrotate' commands.

PR #493, opened 2016-06-11
Fix small error with 'wp core update' command

jessealter:
Looks like there was a stray "www-data" in front of this line


alexfornuto:
Good catch, thanks!

alexfornuto:
[Few style bug fixes PR493](https://trello.com/c/PQuZxiqP/110-few-style-bug-fixes-pr493)

PR #495, opened 2016-06-17
IPv6 Address Assignment 

pbzona:
I added further explanation on the original PR that raised the issue with the /64 pool. Apparently different sized pools do different things, and a big part of the problem had to do with the default routing settings. The first PR was correct, but I tried to give a more comprehensive and complete fix here.

Also updated the Arch section for clarity as the previous steps had some basic errors in how they described the directory. Nothing technical, and I guess not hard to figure out, but important nonetheless. 

The rest is just simple conventions and grammar.


docs/networking/native-ipv6-networking.md
```diff
@@ -80,6 +79,10 @@ While default IPv6 addresses are configured automatically, you will need to stat
 
 ### Debian/Ubuntu
 
+{: .note}
```
alexfornuto: Since this is true for all distributions, I'm moving it out of the Debibuntu section.

docs/networking/native-ipv6-networking.md
```diff
@@ -92,27 +95,48 @@ While default IPv6 addresses are configured automatically, you will need to stat
       auto lo
       iface lo inet loopback
 
-      # The primary network interface (IPv4)
+      # Automatically brings up the default network interface (eth0)
       auto eth0
-      iface eth0 inet dhcp
```
alexfornuto: Why have you removed this line, without re adding it? Will the Linode still get a DHCP-assigned IPv4 address after reboot without it?

docs/networking/native-ipv6-networking.md
```diff
@@ -92,27 +95,48 @@ While default IPv6 addresses are configured automatically, you will need to stat
...
+      allow-hotplug eth0
 
       # IPv6 Address Blocks
-      # You should add an additional block for each IPv6 address you need configured.
-      # You do not need to configure a block for your default IPv6 address, it will be brought up via SLAAC
+      # Add a static block for your default public IPv6 address, and include a directive for its gateway.
```
alexfornuto: Why doesn't your example include the gateway you say it needs?

docs/networking/native-ipv6-networking.md
```diff
@@ -92,27 +95,48 @@ While default IPv6 addresses are configured automatically, you will need to stat
...
+      net.ipv6.conf.all.forwarding=1
+      ~~~
+
+    For addresses within a /116 pool, the above change is not necessary as routing will be automatically configured. Addresses in a /116 pool are routed to other Linodes on your account within the same datacenter.
+
+3.  Restart networking. This command should be performed in [Lish](/docs/networking/using-the-linode-shell-lish), as it will terminate an SSH connection.
 
         ifdown -a && ifup -a
 
 ### CentOS/Fedora
```
alexfornuto: Since you have a section for Fedora21, this section should indicate which versions of Fedora this applies to. Same on Line 174.

alexfornuto: And on Line 216, it should be that version `or above`.

docs/networking/native-ipv6-networking.md
pbzona: Copy/paste error, adding that line back in.

docs/networking/native-ipv6-networking.md
pbzona: I thought it would add confusion to include the IPv4 configuration when this guide is focused on v6. The lines I left in are just for the interface, which IPv6 would need to work. In retrospect, this could add more confusion if some readers removed this section, disabling IPv4. 

Reverting my changes to what the guide originally said. I tested it to make sure it works on both distros.

PR #496, opened 2016-06-19
Update deprecated syntax

thuselem:
Ansible 2.1 reports that using bare variables is deprecated. I updated the code sample with the full variable syntax.


alexfornuto:
Thanks!

PR #499, opened 2016-06-23
Clarified the correct format of the "action" setting.

sevenpointsix:
This may be stating the obvious, but I believe that the "action" setting
value must be enclosed in brackets, and wrapped in %s. For example:

`%(action_mw)s`

The current instructions don't make this clear, in my opinion.


pbzona:
Hi @sevenpointsix, thanks for your contribution. I agree, the syntax isn't quite as clear as it could be, and your proposed changes solve that problem nicely. I'll be merging a few PRs this week and will be sure to include that in this list. Thanks again for pointing this out!

PR #500, opened 2016-06-25
Windows On Linode

ClickSimply:
This is a guide that shows how to install Windows on a Linode VPS.


ClickSimply:
Need to make some more changes, will submit another pull request when it's all cleaned up.

PR #501, opened 2016-06-26
Windows on linode

ClickSimply:


EugeneKay:
This guide uses a VirtualBox donor VM, which is overly complex IMO. The only benefit of performing a KVM-migrated install(my original approach) was the use of Paravirtualization drivers, which was the only ones available in the original Linode KVM Beta. The alternate Full Virt mode is slower, but does not require jumping through quite so many hoops.

I suggest using a "hard disk installer"(eg, a USB stick) directly. dd the entire disk(including partition table!) to a Raw Image for Installation, use Glish to run through Windows install directly on Linode, and remove the Installer from the Profile. I also strongly suggest including a note on EMS for Lish console access.

pbzona:
Hi @ClickSimply 

While this is a cool concept (and one we get asked about more often than you might think), the logistics of testing and updating a Windows guide regularly would make it very difficult to maintain. Additionally, we're unable to provide support for Windows, and generally stay away from publishing guides on proprietary OSes.

Thank you for your submission, but I'm going to have to close this PR. If you're interested in doing a different topic in the future please feel free to reach out. Even though we can't take this guide, it's very well written and you've clearly got some interesting ideas!

dawesi:
"generally stay away from publishing guides on proprietary OSes." #really it's almost like you don't want to take people's money. #hostingfail

PR #507, opened 2016-07-18
Include directive should be at the new line

detailyang:


alexfornuto:
Thanks @detailyang! I created https://github.com/linode/docs/pull/529 which makes a single space adjustment to match indentation alignment.

PR #511, opened 2016-07-22
Restore section may need updated

jeremye77:
Restore section may need updated regarding file system limitations


brewn:
@jeremye77 -

Since your backup includes your original fstab file (with the entry for ext4) Linux will detect it as being an ext4 filesystem. While the backup is being restored as ext3, Linux sees it as ext4 because it's able to mount it as such. 

This is actually something we intend to fix in future iterations of the Backup Service, but at the moment all backups will be restored as ext3.

PR #513, opened 2016-07-24
Update for newest version of Terraria

artemisbot:
ReLogic changed the file order in the Terraria server packages to include all operating system versions. Also changes file names for latest server version v1.3.2.1


cwlinode:
Thank you for bringing this up. This guide has fallen quite out of date and we're rewriting it significantly to bring it back up to speed and streamline things. It should be published in the near future.

PR #514, opened 2016-07-29
missing word in first sentence.

PatrickCoffey:
i have provided a few different options not sure whichone the author was going for.


alexfornuto:
Thanks for bringing this to our attention! Of course, I can't merge this PR as is. If you just want to point something out, feel free to use an issue instead of a PR. I'll spin off of this branch and make the adjustment, so you're included in the history.

PR #515, opened 2016-07-29
Update for Dovecot 2.2.2.

jerrykrinock:
SUMMARY OF MAJOR CHANGES PROPOSED

• Remove references to the old Dovecot certificate and key which no longer exist in Dovecot, and refer instead to the user's "chained" SSL cert and key.  I found this to be necessary in both the Postfix and Dovecot sections of the tutorial.

• In a couple steps which instruct the user to replace entire files, instead just show the lines that are changed.  I went through and carefully found the changes, and there were small in number.  However, the original Dovecot files have many changes (presumably improvements) in comments, so I think it would be better to use the current Dovecot file.

MORE DETAIL ON EACH OF THE PROPOSED CHANGES
- In Postfix step 2, pasting in those 63 lines, in lines 26 and 27, these point to the default legacy dovecot cert and key which do not exist in the current version of Dovecot.  Instead, one must enter the path to your site’s (chained) cert and private key, as explained below in Dovecot step 19.  The Postix documentation does not explain this detail, but it seems to work the same as Dovecot.
- In Postfix step 12, in the example there are only 5 lines “starting with -o” in each of the two groups, but in my original file there are 10 each.  I followed the instructions literally and uncommented all 20 lines, and it worked
- In Dovecot step 2, regarding /etc/dovecot/dovecot.conf, the current version of Dovecot seems to have updated and improved comments compared to the 91-line listing in the tutorial.  The only significant change was adding line 21: protocols = imap pop3 lmtp.  So, I think it would be better to specify this one change rather than replacing all 91 lines.
- In Dovecot step 4, regarding  /etc/dovecot/conf.d/10-mail.conf, again the current version of Dovecot seems to have updated and improved comments compared to the 345-line listing in the tutorial.  This time, there are four differences.  The first two changes,

mail_location = maildir:/var/mail/vhosts/%d/%n
mail_privileged_group = mail

are correctly given in the tutorial.   (The second one is an “uncomment”).

The other two changes are:

~ line 43:
Linode Tutorial says:   namespace {
Dovecot 2.2.22 has:     namespace inbox {

~ line 62:
Linode Tutorial says:   #inbox = no
Dovecot 2.2.22 has:    inbox = yes

Since the tutorial did not mention these, I decided to leave them as in Dovecot 2.2.22.  And it seems to work.
- In Dovecot step 16, I think there is a mistake. Reading the text, I think that the three lines 18-20, given in the tutorial as

inet_listener imap {
  #port = 0
}

should be instead:

inet_listener imap {
  port = 0
}

Probably it does not matter, but I did it my way and it worked.
- In Dovecot step 17, the two files did not exist, and this is expected in the latest version of dovecot.  So, I deleted this step.
- In Dovecot step 19, according to Dovecot’s SSL documentation, http://wiki.dovecot.org/SSL/DovecotConfiguration, and this accepted answer, http://unix.stackexchange.com/questions/146415/specify-certificate-of-ca-in-dovecot, because Dovecot does not separately read intermediate or CA certificates, one must concatenate those after mine into a “chain” file, and set ssl_cert to the chain file.  Also, the instructions should note that the strange-looking less-than character preceding the paths to the cert and key files are required.

### DIFF
--- a/docs/email/postfix/email-with-postfix-dovecot-and-mysql.md
+++ b/docs/email/postfix/email-with-postfix-dovecot-and-mysql.md
@@ -207,7 +207,7 @@ Next, set up Postfix so the server can accept incoming messages for the domains.
 
         cp /etc/postfix/main.cf /etc/postfix/main.cf.orig
 
-2.  Edit the `/etc/postfix/main.cf` file to match the following. Ensure that occurrences of `example.com` are replaced with the domain name. Also, replace `hostname` with the system's hostname on line 44.
+2.  Edit the `/etc/postfix/main.cf` file to match the following, with these changes:(a) Replace occurrences of `example.com` are replaced with your domain name. (b) on line 44, replace `hostname` with the system's hostname.  (c) On lines 26 and 27, enter the indicated paths.  Regarding the certificate file, usually SSL providers will give you two files: one containing your cert and a second file containing one or more certs of the certificate authority.  Unlike Apache, Postfix only reads one cert file, so you must contatenate these two files into a "chained SSL certificate", and enter its path here.  (You will use this same file for Dovecot later.)
 
     {:.file }
     /etc/postfix/main.cf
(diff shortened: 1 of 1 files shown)


jerrykrinock:
Thank you to whomever wrote this guide.  It is straightforward and very well written.  But it was so long that I was amazed when, after making the few changes indicated in this pull request, it actually worked :)

Guaris:
I'm closing this.

PR #519, opened 2016-08-05

piradmin:
Changes reflect the proper method for CA certificate updating in current Linode Ubuntu environment. Previous method would appear to work in some browsers (Chrome,Firefox on desktop) and not others (Chrome mobile). Also failed Thawte installation test until I did it this way.


alexfornuto:
This looks great, thanks @piradmin!

PR #520, opened 2016-08-07
ncsa_auth not found change with basic_ncsa_auth

andhikamaheva:


alexfornuto:
Thanks for the correction. Can you cite a reference to this change? 

Additionally, the extra lines you added to the metadata do not belong, and would break the document if published.

andhikamaheva:
I tried it myself. : D @alexfornuto

PR #522, opened 2016-08-07
Installing and Configuring Samba On Ubuntu Server

YPTanjung:
This is my writing about installing and configuring samba on ubuntu server.


alexfornuto:
Per the latest update to https://github.com/linode/docs/blob/master/CONTRIBUTING.md, we're not currently accepting new submissions. Once the new submission process is in place, we'll update the repo again. Thanks!

PR #523, opened 2016-08-09
changed .gz to .xz

Swashy:
Arch Linux began switching over to .xz in 2010 because its better

https://www.archlinux.org/news/switching-to-xz-compression-for-new-packages/


alexfornuto:
Thanks @Swashy!

PR #534, opened 2016-08-12
Update install-lamp-on-ubuntu-16-04.md

maidentaiwan:
The mysql_secure_installation script is no longer necessary in MySQL 5.7 and above, and can actually cause problems: see https://bugs.launchpad.net/ubuntu/+source/mysql-5.7/+bug/1610574.


EdwardAngert:
Thanks @maidentaiwan!
I've created a duplicate of the PR (https://github.com/linode/docs/pull/547) so that I can merge it right into master (after another tweak or two on our end).
While this PR won't show as merged, the new one will and the master repo will show your contribution.

Thanks again for catching this!
Edward

maidentaiwan:
You're very welcome! I'm glad other people can benefit from the
head-scratching I went through to figure this out.

PR #537, opened 2016-08-17
Update install-gogs-on-debian-8-jessie.md

lmik:
quick fix, wrong file name mentioned


EdwardAngert:
Thanks for catching this @lmik!
We had a PR for a similar issue that we were working on (https://github.com/linode/docs/pull/548), and since we've resolved the issue there, I'm going to close this PR.

Thanks again!
Edward

PR #541, opened 2016-08-25
Updated incorrect language, added helpful image

Swashy:
Language said to click "deploy a distribution" to deploy an image. It is incorrect.
I added an image to make where the link is more clear, and some pointers about deploying deleted/saved images


alexfornuto:
Thanks for the contribution! Your commit doesn't include the new image, you need to add that file to a new commit and add it to your branch.

Swashy:
Whoops! I had uploaded it to docs/assets, but I must have messed up somewhere. here it is.

alexfornuto:
Thanks! I'll look into removing the old image, assuming it's not used anywhere else, and getting this merged next week.

Swashy:
Alex, the original image is fine to keep, its a highlight of the drop down menu. With my additional image, it shows you where the link is to get to that drop down menu is(I find the link relatively small and unnoticeable UI-wise)

alexfornuto:
Thanks for the clarification!

PR #545, opened 2016-08-31

rlamfink:
SSLCACertificateFile must be set to /etc/ssl/certs/ca-certificates.crt, not localcerts. This is to match the reference on line 53.


alexfornuto:
Thanks, this issue has already been resolved!

PR #550, opened 2016-09-08
CONFIG_BLK_DEV_SD is needed also

tomrittervg:
@akerl @jessealter Helped me debug a problem with a Xen -> KVM migration and missing this kernel config option prevented the boot.


alexfornuto:
Thanks guys!

PR #552, opened 2016-09-10
Fix iptables usage

dwfreed:
Somebody deserves to be drawn and quartered for this, but I can't tell who, because these problems predate this repo.  **ALWAYS** **ALWAYS** **ALWAYS** use `-v` when doing an `iptables -L`.  The rules are basically useless otherwise.  `-n` is also a good idea, because DNS resolution is slow and can be confusing.


alexfornuto:
Thanks @dwfreed

PR #553, opened 2016-09-11
add missing jail options

tkimnguyen:
Without an explicit line for ignoreregex, fail2ban-client reload complains.
Without a port line, fail2ban-client reload fails.


alexfornuto:
Thanks!

PR #554, opened 2016-09-11
Securing your server: save modified firewall rules

wturrell:
Debian/Ubuntu - explain how to save modified iptables rules.

The instructions later (for adding, editing, deleting rules) state: 
"To accomplish this, revisit the area above for your distro to 
save your iptables edits so they’re loaded on reboots."

It may not be immediately apparent how to do this in Debian.


cwlinode:
Thank you for bringing this up. The firewall section of this guide has been broken off into a more manageable set of guides for iptables, firewalld and ufw. An explanation for reloading iptables-persistent was added to the iptables guide.

PR #557, opened 2016-09-13
updated fail2ban.local steps to fix #518

### LINKED ISSUE #518: not recommended recommendations in using-fail2ban-for-security.md
in fail2ban we do not recommend to just 'cp jail.conf jail.local' and do modifications.  .local files should provide only customizations, overriding settings in jail.conf.  Please adjust the page accordingly.

alexfornuto:


docs/security/using-fail2ban-for-security.md
```diff
@@ -126,11 +126,14 @@ Fail2ban reads its configuration files so that all `.conf` files are read first
 
         cd /etc/fail2ban
 
-2.  Copy `fail2ban.conf` to `fail2ban.local`:
+2.  The file `fail2ban.conf` contains the default configuration variables for Fail2ban logging, the socket used to communicate with the daemon, and the location of the PID file. Changes should be made in a separate file, `fail2ban.local`, which overrides `fail2ban.conf`. If you want, you can copy `fail2ban.conf` to `fail2ban.conf`, commenting out all variables, and then uncomment only the options you want to modify:
```
yarikoptic: may be "which overrides" -> "definitions in which override"

docs/security/using-fail2ban-for-security.md
```diff
@@ -126,11 +126,14 @@ Fail2ban reads its configuration files so that all `.conf` files are read first
...
 
         cp fail2ban.conf fail2ban.local
+        sed -i 's/^/#/g' fail2ban.local
```
yarikoptic: kinda would work if people wouldn't also forget to uncomment section names ;-)  and should provide remedy for the issue of stale configurations propagating across releases
would I use such approach myself?  probably not since it is hard to see what I am actually changing without running some 'grep' command, people would tend to uncomment more than necessary to achieve the goal, etc

docs/security/using-fail2ban-for-security.md
```diff
@@ -126,11 +126,14 @@ Fail2ban reads its configuration files so that all `.conf` files are read first
...
 
-3.  Open `fail2ban.local` in your chosen text editor. This file contains configuration for Fail2ban logging, the socket used to communicate with the daemon, and the location of the PID file. The values that can be changed within the `fail2ban.local` file are as follows:
+    Otherwise, create a blank `fail2ban.local` file, and manually add the options you wish to modify.
```
yarikoptic: "options" -> "options (with corresponding secions)" ?

docs/security/using-fail2ban-for-security.md
```diff
@@ -126,11 +126,14 @@ Fail2ban reads its configuration files so that all `.conf` files are read first
...
+
+3.  In `fail2ban.local` add or uncomment and edit the values to match your desired configuration. The values that can be changed within the `fail2ban.local` file are:
 
     -   `loglevel`: The level of detail that Fail2ban's logs provide can be set to 1 (error), 2 (warn), 3 (info), or 4 (debug).
     -   `logtarget`: Logs actions into a specific file. The default value of `/var/log/fail2ban.log` puts all logging into the defined file. Alternately, you can change the value to STDOUT, which will output any data; STDERR, which will output any errors; SYSLOG, which is message-based logging; and FILE, which outputs to a file.
```
yarikoptic: below, not covered by this diff, you have also section on 'jail.conf' which still starts with 'cp jail.conf jail.local' ;)

docs/security/using-fail2ban-for-security.md
pbzona: @yarikoptic Here's the above simplified into one command, matching only the definitions and not the section name:

    sed 's/\(^[[:alpha:]]\)/# \1/' fail2ban.conf | sudo tee fail2ban.local 1&> /dev/null

Could you clarify what you mean by the issue of stale configs across releases? I'm not sure that needs to be addressed here since users should update (or at least check) their .local file when they upgrade to a new release. Any changes they've already made to an outdated .local file won't easily be updated without a custom script or something that goes beyond the scope of a one-line command.  I might be misunderstanding your point though.

docs/security/using-fail2ban-for-security.md
alexfornuto: Check out L143 now.

alexfornuto:
@yarikoptic @pbzona took a pass on this and I believe he address most or all of your concerns. If you throw a 👍  on this I'll get it live on Monday.

yarikoptic:
+1 even though, I personally prefer lean jail.local with only what jails I want to have enabled eg

```
[sshd]
enabled = true

[apache-auth]
enabled =true
```

Makes it clear what is enabled and what not without digging through commented out sections. But technically both are identical, thus +1

alexfornuto:
I personally agree with you, but as an introductory guide, we want it be easy to quickly configure and understand. As the reader becomes more familiar, they can prune the document as desired.

PR #563, opened 2016-09-20
Fix typo

paulotfilho:
Typo on "If you want, you can copy `fail2ban.conf` to `fail2ban.conf`". Should be "If you want, you can copy `fail2ban.conf` to `fail2ban.local`"


alexfornuto:
Thanks @paulotfilho!

PR #566, opened 2016-09-25
Correcting packages to install for PHP

alexfinnarn:
When I tried the current line, I got an error, but this package combination worked.

### DIFF
--- a/docs/websites/hosting-a-website.md
+++ b/docs/websites/hosting-a-website.md
@@ -275,13 +275,9 @@ PHP is a general-purpose scripting language that allows you to produce dynamic a
 
 ### Install PHP
 
-1.  Install the base PHP package:
+1.  Install the PHP package with MySQL support:
 
-        sudo apt-get install php5 php-pear
-
-2.  Add the MySQL support extension for PHP:
-
-        sudo apt-get install php5-mysql
+        sudo apt-get install php libapache2-mod-php php-mysql
 
 ### Optimize PHP for a Linode 2GB
 


alexfornuto:
[Correcting packages to install for PHP](https://trello.com/c/BH80mLNQ/177-correcting-packages-to-install-for-php)

pbzona:
Hi @alexfinnarn, thanks for bringing this to our attention. What was the error you received when you followed the guide, and could you provide some more information about the system you tried it on? At a glance, this seems like it should work, but I'd like to test it for myself to see if I can replicate what you saw.

alexfinnarn:
Sorry, this was a shitty ticket...I have no idea what the error message was or could be. I did use a pre-packaged image, 16.04 LTS Profile (Latest 64 bit (4.8.6-x86_64-linode78))...but that's all I can tell you unfortunately.

pbzona:
@alexfinnarn No worries, I'll look into it either way. I think there are a few PHP packages on Ubuntu 16.04 that default to PHP 7, and I've seen some compatibility issues because of that. I'll run some tests using the guide and see if I'm able to nail it down to something specific, and if so, we can make updates as needed. I should have an update for you next week at the latest.

pbzona:
Hey @alexfinnarn, the issue was actually simpler than I thought - the guide isn't compatible with Ubuntu 16.04, but that fact is buried in the note at the beginning:

https://linode.com/docs/websites/hosting-a-website

The guide was updated a few days before you made this PR, so you may have been viewing a cached version that didn't include the note (even if not, it's very easy to miss). For Ubuntu 16.04 installations, you'll want to check out this guide:

https://linode.com/docs/websites/lamp/install-lamp-on-ubuntu-16-04

In that one, we specify the same PHP packages you do, just for version 7 (it would work with `php` packages rather than `php7.0` as well - I was wrong earlier, `php` defaults to version 5). I'm going to close this PR since we essentially already have the information you submitted, but this brings up something else we'll want to look into - visibility. The note excluding Ubuntu 16.04 is easy to miss, so I'll bring that issue to the team and see if we can come up with a way to make this clearer.

Thanks again for your submission. Please let us know if you have any other suggestions to improve our docs!

PR #567, opened 2016-10-01
Changes in /etc/postfix/main.cf

mcmidget:
I had to change these settings in order to setup a working DKIM+Postfix under Debian jessie.

Replace:
        smtpd_milters = local:/opendkim/opendkim.sock
        non_smtpd_milters = local:/opendkim/opendkim.sock

With:
        smtpd_milters = unix:opendkim/opendkim.sock
        non_smtpd_milters = unix:opendkim/opendkim.sock

### DIFF
--- a/docs/email/postfix/configure-spf-and-dkim-in-postfix-on-debian-8.md
+++ b/docs/email/postfix/configure-spf-and-dkim-in-postfix-on-debian-8.md
@@ -356,8 +356,8 @@ If everything is OK you shouldn't get any output. If you want to see more inform
         milter_default_action = accept
         # Postfix ≥ 2.6 milter_protocol = 6, Postfix ≤ 2.5 milter_protocol = 2
         milter_protocol = 6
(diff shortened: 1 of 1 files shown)


alexfornuto:
@tknarr any input on this?

tknarr:
I'm running Jessie on my systems and the existing settings are working fine. In fact they're required because Postfix is running chrooted at /var/spool/postfix. A quick check would be the SOCKET setting in /etc/defaults/opendkim. In mine it's the default path:

    SOCKET="local:/var/spool/postfix/opendkim/opendkim.sock"

which lines up with the existing Postfix settings. @mcmidget what's the SOCK setting for your installation?

mcmidget:
Mine is SOCKET="local:/var/spool/postfix/opendkim/opendkim.sock" too

cwlinode:
Closing this because there don't seem to be any further issues. We'll revisit it if this comes up again in the future.

PR #574, opened 2016-10-09
Fix Gentoo command to install Git

cfackler:
Fix Portage --ask flag and git is now in the dev-vcs category.


alexfornuto:
Thanks!

PR #575, opened 2016-10-10
Workaround for docker installation bug

dmcsorley:
Error (installation hangs forever) occurs on at least docker 1.11-1.12.1 and ubuntu 16.04.


pbzona:
Hi @dmcsorley, thanks for your submission. I've been having a nighmare of a time with this exact issue on some Docker installations recently, I wish I'd known about this solution sooner! We should be merging a batch of updates in the near future, and I'll be sure to include this when we publish them. Thanks again for your contribution!

PR #584, opened 2016-10-30
Update use-public-key-authentication-with-ssh.md

somenxavier:
Using ssh-copy-id from Arch [https://wiki.archlinux.org/index.php/SSH_keys#Copying_the_public_key_to_the_remote_server]


pbzona:
Hi @somenxavier, thanks for your submission. The `ssh-copy-id` command isn't available on every OS by default, and this guide is intended to include instructions that will work on a variety of local systems. I'd be glad to look into it further if there are any specific benefits to using `ssh-copy-id` rather than `scp` - are there any I should be aware of?

somenxavier:
OK. Thanks for the info.

PR #583, opened 2016-10-28
SSH Hardening Guide proposal

damasosanoja:
## **IMPORTANT: this pull request is the same #436**

**_I accidentally deleted my fork on GitHub after losing my local copy due to human error. Something to laugh TGIF ..._**

> @alexfornuto  commented 2 days ago:
> Sounds good @damasosanoja, I suggest you start by filling out the introduction, and the section headers for us to review.

Greetings @alexfornuto, here is the first draft. Includes:
- A new "SEO friendly" Title proposal: **7 Easy Ways to improve SSH Server security**
- Brief article introduction.
- First step draft (with attached image for easier understanding).
- Section headers proposal (steps).
- Last step brief introduction with attached image. About this step, you will notice has a different approach because is not "really" part of SSH, and also because is much more extensive than the "main" article. Not sure if this is the appropriate angle to introduce the topic, but it's a start.

PR #585, opened 2016-11-01
Ssh hardening guide draft

damasosanoja:
Hi @alexfornuto 

New pull request again, I learned my lesson against Murphy's Law today. This comes from #583

**Includes:**

- A new "SEO friendly" Title proposal: **7 Easy Ways to improve SSH Server security**
- Brief article introduction.
- First step draft (with attached image for easier understanding).
- Section headers proposal (steps).
- Last step brief introduction with attached image. About this step, you will notice has a different approach because is not "really" part of SSH, and also because is much more extensive than the "main" article. Not sure if this is the appropriate angle to introduce the topic, but it's a start.


damasosanoja:
Hi, any updates on this guide?

damasosanoja:
Hi @pbzona is there any news on this guide proposal, thanks in advance.

alexfornuto:
Hi @damasosanoja. First off, while we like when our guides rank well on Google, we'll do it naturally and without the clickbait-y titles. But I like where your head is at!

Your outline looks good, but I have one question. I see that you updated the proposed "port knocking" techniques you'd cover based on the discussion with @akerl in your first PR. My question for you (and him if he's so inclined) is if those techniques are still considered "port knocking" or if there's a more apt description.

damasosanoja:
HI @alexfornuto you made a good question.
From developers point of view **fwknop** [http://www.cipherdyne.org/fwknop/](url) and **knockknock** [https://moxie.org/software/knockknock/](url) are indeed port knocking implementations.
If you trust wikipedia it says:
> A variant called single packet authorization exists, where only a single "knock" is needed, consisting of an encrypted packet

That said, both **knockknock** and **fwknop** use encrypted Single Packet Authorization which is the main advantage/difference with original port knocking implementation. So an alternative description would be SPA methods.
I would like to hear @akerl perspective on this too =)

akerl:
"It's complicated".

It all depends on how you define port knocking. If port knocking is the generic concept of sending specific traffic to a port, and using that to trigger opening a firewall hole, Single Packet Auth (SPA) is a subset of port knocking that uses cryptographically secure request packets. If port knocking is the specific concept of sending empty or arbitrary packets to a sequence of ports that causes a firewall hole to be opened, SPA is a separate concept entirely.

I'm inclined to take the latter stance, partly because it feels more clear about what strategy is being employed and partly because "port knocking" carries with it a lot of negative connotation from the security side (since the non-SPA approach is vulnerable to brute force and replay attacks and a host of other nastiness).

damasosanoja:
Totally agree with @akerl input, specially the "negative connotation" of traditional port knocking. What about explaining traditional port knocking and then introduce SPA techniques in a separate section something like Step 8: Port Knocking, Step 9: Single Packet Authorization. 
An alternative would be to skip PK section but I think is a good idea to illustrate the concept and mention its disadvantages before talking about SPA.
What do you think @alexfornuto ?

alexfornuto:
@damasosanoja sorry for the delay. That question should be directed now at @pbzona, @EdwardAngert @cwlinode or @brewn.

damasosanoja:
Excellent thanks @alexfornuto

damasosanoja:
Hi @pbzona  @EdwardAngert @cwlinode @brewn , there is any update on this topic? Thanks in advance.

pbzona:
Hi @damasosanoja 

Thank you again for your patience. We're wrapping things up with your Odoo guide and would like to move on to this one as soon as possible. I notice you have an outline, which looks great - how much time do you think you'd need to fill that out into a complete guide? Just asking so we can plan our publication schedule accordingly.

As far as the port knocking section, it's a tricky topic. On one hand, what you're laying out is useful from a security perspective. On the other, I'm not sure we want to get into a topic that has such negative connotations...at least not until I can get our team together to have a longer discussion about it.

Here's what I propose: leave the port-knocking section off for now, and we can potentially do a future guide on SPA (and its relationship to port knocking) since it's a complex topic and the rest of this guide is substantial enough to stand on its own. Sound fair?

damasosanoja:
Hi @pbzona 
I agree with your point of view about SPA. It's a content-rich topic that can be addressed from several perspectives, some positive, some not. Writing a separate article is an excellent solution.

That said, I'm not sure if the proposed SSH Hardening guide (without SPA) would be long enough to be considered "an article", maybe it can be a kind of append or section of the current **Securing Your Server** guide? 

What about this: I can finish the SSH Hardening guide this weekend then you can review it and decide which is the best approach, what do you think?

pbzona:
@damasosanoja That's perfect, I'll take a look as soon as it's ready and we can figure something out. Thank you!

damasosanoja:
Hi @pbzona 
The first draft in ready for review, best regards.

pbzona:
Thanks @damasosanoja!

I'm going to read through and double check that everything works as expected. Should have a PR for you with some suggestions in the next day or two.

damasosanoja:
Excellent!

pbzona:
@damasosanoja Sorry for the delay! I've made a pull request to your branch with some changes and comments. Once you have a chance to review, please merge those and we'll get one of my colleagues to do a final review before publishing. Thanks!

pbzona:
@damasosanoja I'm going to have a graphic done for the header and we should be all set to publish. Please keep an eye out for a PR early next week!

damasosanoja:
@pbzona 
A graphic! Can't wait! Excellent thanks!

PR #590, opened 2016-11-03
bugfixes for fail2ban config files

briancappello:
* do not create a junk file named ``1``
* make ``jail.local`` syntactically valid so that fail2ban will start


pbzona:
Nice catch @briancappello! I've noticed that issue with the `jail.local` file but haven't been able to find time to work out the regex. I tried it out and this does the trick. I'll be going through over the next few days and merging some community submitted updates, and will be sure to include this in that round. Thank you!

PR #593, opened 2016-11-12
update pgAdmin links

stvnjacobs:
pgAdmin 3 is no longer supported. Both guides have been updated to use their respective links to download pgAdmin 4.

I also would consider linking to the root pgAdmin `/download/` page, so the reader can decide for themselves, but that is up to all of you.  Most Windows users will probably be fine on pgAdmin 4 (minimum requirement of Windows 7), but pgAdmin 4 for macOS requires 10.10, and some may not be up to date.

I avoided updating the wording to use 4 throughout until someone else weighs in, but would be happy to update this pull request.


stvnjacobs:
After all that, I didn't mention the real reason I started this pull request.  The Mac download link is dead (404).

pbzona:
Nice catch, @stvnjacobs. I'll be reviewing/merging PRs this week, and will be sure to include this. Thanks!

PR #565, opened 2016-09-24
[mysql] didn't work but [mysqld] loaded fine

alexfinnarn:

PR #597, opened 2016-11-18
error in block insert in my.cnf

karliatto:
the block should start with `[mysqld]` instead of `[mysql]`


pbzona:
Hi @karliatto, thanks for your contribution. This issue is addressed in https://github.com/linode/docs/pull/565. It seems you are correct, but since it was brought up in a previous PR I'm going to close this one. Thank you again for pointing this out!

PR #598, opened 2016-11-19
Make SSH is using PAM

LouisMT:
If SSH isn't using PAM, the SSH client will just disconnect


pbzona:
@LouisMT The`UsePAM` setting is enabled by default in the distros covered by these guides. This is a good troubleshooting step, but I'm not sure there are many common scenarios in which a user might find it has been disabled. Could you provide more info regarding the need for such a note?

LouisMT:
@pbzona I disabled it sometime to completely disable password logins via SSH. I know it isn't necessary, but many tutorials suggest disabling PAM. Google "disable SSH password login".

pbzona:
Hey, @LouisMT, we haven't forgotten about you. I think you raise a good point here, and I agree the `UsePAM` directive should be explicitly stated in case others have been playing with their SSH settings prior to following our 2FA guides. I'll be reviewing and merging a number of PRs this week, and will be sure to add this to the list. Thanks for your contribution!

PR #602, opened 2016-11-21
Update to conform to Debian Kernel Handbook

markybob:
https://kernel-handbook.alioth.debian.org/ch-common-tasks.html#s-kernel-org-package


cwlinode:
Thank you for updating this!

PR #603, opened 2016-11-22
fix for all_video.mod error

markybob:


cwlinode:
Thank you for making this pull request. We've recently updated the distro-supplied kernel guide with a Grub configuration that no longer results in this error.

PR #608, opened 2016-12-02
Syntax corrections and more checks

da2x:
* Fixed syntax errors.
* Check that everything worked before exiting.

### DIFF
--- a/docs/websites/cms/high-availability-wordpress.md
+++ b/docs/websites/cms/high-availability-wordpress.md
@@ -3,11 +3,12 @@ author:
   name: James Stewart
   email: jstewart@linode.com
 description: 'Configuring a highly available WordPress installation.'
-keywords: 'wordpress,mysql,replication,master-master,high availability'
+keywords: 'wordpress,mysql,mariadb,replication,master-master,high availability'
 license: '[CC BY-ND 4.0](https://creativecommons.org/licenses/by-nd/4.0)'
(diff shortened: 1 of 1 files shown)


pbzona:
Hi @da2x 

Thanks for your submission. I'm going to test your changes today just to be sure everything is working as expected - I should have an update for you shortly.

da2x:
I forgot to include syntax documentation:
https://mariadb.com/kb/en/mariadb/start-slave/
https://mariadb.com/kb/en/mariadb/stop-slave/
https://mariadb.com/kb/en/mariadb/show-slave-status/

This isn’t really a _MariaDB thing_ as the syntax here is the same for MySQL. MariaDB ships by default pretty much everywhere now as a drop-in replacement for MySQL.

There, information previously only held in my brain is now included here for the record.

pbzona:
@da2x The guide seems to have been tested on Ubuntu 14.04 and Debian 7, which would include MySQL (as opposed to MariaDB). I ran through the first few sections and it seems to be working as expected; what distro were you using when you encountered this?

pbzona:
@da2x I'm going to close this as the syntax appears to be working both ways. While the documentation you've provided is helpful, the guide is specifically meant for distros that ship with MySQL. If you can provide more detail on why you believe this to be an error (not just a difference in syntax) I'll be glad to revisit it in the future.

PR #610, opened 2016-12-06
clarifies use of "open_file_cache_min_uses"

uschtwill:


pbzona:
Hi @uschtwill, thanks for your contribution. I agree that the current wording isn't as clear as it could be, and your suggestion is definitely an improvement. I'll mark this to be merged along with several other PRs this week. Thanks again!

PR #591, opened 2016-11-09
Added step to install PHP GD library for image processing

pbzona:
I discovered this incidentally while working on a personal site. Without the PHP GD library, users can't modify uploaded images (i.e., cropping to create a favicon, etc.). It's likely this change applies to other docs, but I made my initial commit based on the environment I found it in.

PR #615, opened 2016-12-09
Include instructions for PHP extensions like php-xmlrpc

jakelevirne:
Without these instructions, Jetpack installation and use of the Wordpress mobile app are not possible


pbzona:
Hi @jakelevirne, thanks for your contribution. I like the idea of including optional PHP extensions, and made a similar, but less comprehensive, PR not long ago (https://github.com/linode/docs/pull/591). Would you mind adding a list of what the various extensions do, so that users can figure out which ones they'll need to install for their system?

pbzona:
@jakelevirne Thanks for your patience. I'll be going through over the next few days to merge some PRs, and I'll be sure to include this. Thanks again for your contribution!

PR #617, opened 2016-12-10
Update disk-images-and-configuration-profiles.md

RodrigoJimmy:
Just fixing to reflect the right text string on link.


pbzona:
Hi @RodrigoJimmy, thanks for your contribution. It does appear we have text from an outdated version of the manager, nice catch! I'll be merging PRs this week, so I'll be sure this gets included. Thanks again!

PR #619, opened 2016-12-11
Fixed links to the two 'Obtain a Commercially Signed SSL Certificate'…

edruder:
… pages


pbzona:
Nice find @edruder! Normally I would be inclined to close without merging since this is a deprecated guide, but since the reason it was deprecated was the splitting into two new docs, I think you raise a good point. I'll be reviewing and mergind PRs over the next few days and will be sure to include this. Thanks for your contribution!

PR #620, opened 2016-12-15
generated dkim entry incorrect syntax for sha-256

jasonblewis:
when you follow the instructions, the text record has an incorrect hash entry as h=rsa-sha256, modifying it in the record to h=SHA56 fixes the issue.

without this fix, the `opendkim-testkey -d organictrader.com.au -s 201612` test failes with:
```
opendkim-testkey: unknown hash 'rsa-sha256'
```

### DIFF
--- a/docs/email/postfix/configure-spf-and-dkim-in-postfix-on-debian-8.md
+++ b/docs/email/postfix/configure-spf-and-dkim-in-postfix-on-debian-8.md
@@ -295,12 +295,12 @@ example.txt
 
     ~~~
 
-The value inside the parentheses is what you want. Select and copy the entire region from (but not including) the double-quote before `v=DKIM1` on up to (but not including) the final double-quote before the closing parentheses. Then edit out the double-quotes within the copied text and the whitespace between them. Also change `h=rsa-sha256` to `h=SHA256`. From the above file the result would be:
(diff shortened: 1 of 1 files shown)


alexfornuto:
@tknarr would you care to weigh in on this?

tknarr:
That part of the text's generated by the OpenDKIM tools themselves. When I run `opendkim-testkey` on a record that has `k=rsa` in it I get a successful result:

    root@arachnae:/etc/opendkim/keys# opendkim-testkey -d silverglass.org -k ./silverglass.201610.key -s 201610 -vv
    opendkim-testkey: using default configfile /etc/opendkim.conf
    opendkim-testkey: key loaded from ./silverglass.201610.key
    opendkim-testkey: checking key '201610._domainkey.silverglass.org'
    opendkim-testkey: key not secure
    root@arachnae:/etc/opendkim/keys# exit

I'm not sure where @jasonblewis went from the `k` tag specifying the key algorithm to the `h` tag specifying the list of acceptable hash algorithms. The default if the `h` tag isn't present is all valid hash algorithms for the key algorithm, which is usually what we'd want.

cwlinode:
jasonblewis, thank you for pointing this out. It seems that OpenDKIM generates the `h=rsa-sha256` option along with `k=rsa`.

rejon:
> when you follow the instructions, the text record has an incorrect hash entry as h=rsa-sha256, modifying it in the record to h=SHA56 fixes the issue.
> 
> without this fix, the `opendkim-testkey -d organictrader.com.au -s 201612` test failes with:
> 
> ```
> opendkim-testkey: unknown hash 'rsa-sha256'
> ```

think you mean SHA256 not SHA56, so in total  chage:

 h=rsa-sha256 

to

 h=SHA256

PR #621, opened 2016-12-15
Update install-php-fpm-and-apache-on-debian-8.md

jdzeller:
Line 199 reads Alias /php5-fcgi /usr/lib/cgi-bin/php5-fcgi-site1.net and should read Alias /php5-fcgi /usr/lib/cgi-bin/php5-fcgi-site1.com to match the first path in line 200. After making this change in my own configuration it began working as it should.


pbzona:
Nice catch @jdzeller! I'll be merging PRs throughout the week and will mark this to be included. Thanks for your contribution!

PR #622, opened 2016-12-16
changed contribution links

omps:


pbzona:
Hi @omps - thanks for your contribution. The syntax here is actually correct. The site accepts relative file paths for other docs in our repo, and that's been our convention when linking to them. Additionally, the Gitlab guide's URL is aliased to the guide, so it's not an issue. As such I'll be closing this PR, but nice eye for detail!

PR #627, opened 2017-01-13
SSL settings modified and minor typos corrected

Fred-Zweig:
A reader of [my guide](https://linode.com/docs/websites/varnish/use-varnish-and-nginx-to-serve-wordpress-over-ssl-and-http-on-debian-8) pointed out that he used the settings in the guide to serve a single https site and he received the following error: 
```
400 Bad Request
The plain HTTP request was sent to HTTPS port
nginx/1.10.0 
```
After closer examination and a lot of testing I found out that the above error appears only when one configures Nginx to serve a single https site, without any additional http site. When Nginx is configured to serve at least one https site and one http site, the error doesn't appear. This explains why this error didn't show up during my tests or during Linode's tests. 

Anyway, this error reveals an anomaly in the SSL settings of Nginx. If the SSL directives (`ssl on; ssl_certificate ... ;`, etc. ) are specified only in the first server block, which is the server block for port 443, the error disappears. As the guide stands now, the SSL directives are included in both server blocks: the 443 server block and the 8080 server block. The fact is that when the 8080 server block comes into play, the traffic is already decrypted, so, the SSL directives don't play any role in this block. I remember that I placed the SSL directives in the 8080 server block because it seemed to solve some odd behavior of a test site, but the cause of the problems I had were elsewhere. So, since the SSL directives don't play any role in the second server block of the https site, and since they even cause the above mentioned error when Nginx serves a single https site, I removed them in this pull request. I also corrected a few minor typos, and the scheme picture, because it contained "Varnish replies **to to** Nginx" and had one-sense arrows from clients to Varnish/Nginx instead of double-sense arrows, to indicate that data travels both ways.

I closed my previous pull request on this subject because of a conflict, and opened this new PR.


Fred-Zweig:
@alexfornuto  @pbzona  Please check this pull request and include these corrections in the published guide. While it is unfortunate that the defective SSL settings got into the final version of the guide, the problem can be easily solved by merging this pull request. I'm glad that the error has been discovered rather early after publication.

pbzona:
@Fred-Zweig Thanks for the update. We've had a lot on our plates recently, but I should be able to get to testing this at some point today. Everything looks fine based on my knowledge of nginx, but we do need to double check everything before merging. I'll get back to you today or tomorrow so we can resolve this as soon as possible.

Fred-Zweig:
@pbzona  Thanks a lot!

pbzona:
Hey @Fred-Zweig thanks again for this. I've gone through and verified that everything is working correctly, so I'll include this in the next batch of merges (probably in the next few days or so). Thank you!

alexfornuto:
bitmoji

PR #629, opened 2017-01-16
Replace multiple instances of `-0` with correct `-O`

aularon:
Zero was used instead of capital letter O!


pbzona:
Hey @aularon, thanks for bringing this to our attention! We should be merging and publishing a batch of updates in the near future, and I'll be sure to include this. Thanks again for your contribution!

alexfornuto:
@EdwardAngert 😊

PR #631, opened 2017-01-23
Minor bug fix update for  configure-spf-and-dkim-in-postfix-on-debian-8.md

mullermn:
Section 8 of 'Configure OpenDKIM' contains (what I believe is) an error; the -h parameter should be sha256, not rsa-sha256. Using the latter causes signed mails to fail validation using the tests at check-auth@verifier.port25.com and appmaildev.com.

I tracked this down thanks to a clue in http://arstechnica.com/civis/viewtopic.php?t=1238575&start=120 (2nd post) which also says this error is present in much of the documentation around the internet.


alexfornuto:
I believe this is a duplicate of #620 

bitmoji

mullermn:
Ah yes, that looks like the same issue.

alexfornuto:
bitmoji

PR #634, opened 2017-02-07
Steps to add storage space needs to be more obvious

Swashy:
Inserted a step #7 to point out the need for adding un-used storage space to a disk on a Linode. Also added an image demonstrating how to do this.


cwlinode:
Thank you for making these additions!

PR #642, opened 2017-03-14
Add GRUB_CMDLINE_LINUX parameter to GRUB config

sypherspace:
Without this entry, the Lish console is blank after initramfs loads. No output or interaction is possible.

Tested on Debian 8.7.


sypherspace:
Ignore me, I just noticed that that information was simply relocated elsewhere in the guide.

PR #644, opened 2017-03-18
note about wildcard records underneath the "add a record" section.

Swashy:
if you add a wildcard subdomain under a zone like "*.subdomain, then it fails with an invalid message. A record of just '*' is allowed however, so you need to make a new zone for this.


brewn:
@Swashy 

Isn't the issue here that our nameservers don't allow nested subdomains of any kind, not just wildcards? If so, it's probably better to expand this to include all A records on subdomains.

brewn:
Fixed here:

https://github.com/linode/docs/pull/657

PR #647, opened 2017-03-23
Docker seems to use OverlayFS(overlay2) by default now when available.

shinji257:
On a recent installation it was noted that while it still gives the warning regarding aufs missing it now uses OverlayFS (overlay2) as the storage driver instead of DeviceMapper.


cwlinode:
Thank you for the update.

PR #651, opened 2017-03-30
copy edit

brewn:


brewn:
This should sit until the updated graphics are ready.

PR #658, opened 2017-04-01
Install & configure Oracle XE 11g on Ubuntu 14

tallona:
Needed to do this recently and found there was no documentation on it. Documented it as I went along and added some of the issues I came across and how I fixed them.


cwlinode:
@tallona Thank you for your guide submission.

We are restructuring the docs library and would appreciate if you made a new pr in about a month's time so we can cleanly merge your guide. Also please see our formatting guide and complete the header and Before You Begin sections for your new pr.
https://linode.com/docs/linode-writers-formatting-guide

Thanks!

PR #660, opened 2017-04-05
OpenVPN Guide: Add step to install easy-rsa package

janmuennich:


janmuennich:
Closing again, on stable jessie the package is already included

PR #661, opened 2017-04-06
xvd[ab] --> sd[ab]

Swashy:
xvd[ab] --> sd[ab]


cwlinode:
Thank you for this fix!

PR #663, opened 2017-04-18
Update macOs cache flush command

schottey:
The command to flush the macOS DNS cache was not updated beyond Yosemite,
and in fact changed back to the old one in 10.10.4.


kcraig34:
Schottey -

Thanks for the recommended revision. We've merged it with the master branch.

PR #666, opened 2017-04-25

bestshop24h:
1)for letsencrypt
2)if create more than one ssl under one IP


kcraig34:
bestshop24h-
Thanks for alerting us to these proposed revisions. 
We'll make appropriate changes after additional review.

PR #669, opened 2017-05-02
update team number

alexfornuto:


EdwardAngert:
Thanks @alexfornuto!
I'm going to make a minor tweak to this and change the actual number to just "small" to keep things accurate and non-gendered.

PR #674, opened 2017-05-11
A typo change in introduction-to-linux-concepts.md

SimpleBeat:
A minor typo (bur instead of but), and a word change (yet instead of but).


kcraig34:
Thanks for the keen copy edit eyes, SimpleBeat.

PR #675, opened 2017-05-12
fix number of connections

deric:
It wouldn't make much sense to have higher number of connections than number of opened files (look like a typo).


kcraig34:
Deric -

Thanks for pointing out the typo. We'll make the recommended change.
Let us know if you find others in the future.

kcraig34:
Select revisions made by Guaris.

PR #677, opened 2017-05-21
Update for php 7 LAMP server

danielpflood:
Updated php installation. Taken from https://wiki.archlinux.org/index.php/Apache_HTTP_Server


Guaris:
Thanks for noticing! We've merged your PR!

PR #678, opened 2017-05-23
Formatting correction in Advanced SSH Security guide

pbzona:
Fixed a bug in the formatting that caused the file excerpt in the "Custom Rules Example" to not render properly.


kcraig34:
Great find, PZona. 
Thanks.
Synching, merging already needed today. Yay!

PR #690, opened 2017-06-06
Debian 8 upgrade netfilter-persistent fix

TheJosh:
The `sed` command to fix the issue with `/etc/modules` file was editing the wrong file.


Guaris:
Thanks for submitting a PR! I'll have the issue resolved by the end of the day!

PR #699, opened 2017-06-15
Avoid deprecation message when installing node

daniquilez:


Guaris:
I'm not sure how I missed this PR. I'll test it out this week and see if I can merge it soon after. Thanks for submitting this to us!

GuessWhoSamFoo:
The user would be better off explicitly seeing the warnings for attempting to install nodejs 0.10 (current recommended at v4) unless there is a strong justification.

PR #713, opened 2017-06-23
links to pricing page instead of listing price

EdwardAngert:
(per recommendation) PRing this for when the information is on /pricing


GuessWhoSamFoo:
What's the status on this? The pricing page does not address block storage yet.

Also currently free in beta:
https://forum.linode.com/viewtopic.php?f=26&t=14906

PR #716, opened 2017-06-26
copy edit to ntopng

EdwardAngert:
renamed the file to match title and renamed the images and their alt text for better accessibility


docs/networking/diagnostics/deploy-ntopng-on-debian-8.md
```diff
@@ -0,0 +1,180 @@
...
+
+3.  Verify that the `tcp-segmentation-offload` section in the output matches the following:
+
+        tcp-segmentation-offload: off
+                tx-tcp-segmentation: off
+                tx-tcp-ecn-segmentation: off
+                tx-tcp-mangleid-segmentation: off
+                tx-tcp6-segmentation: off
+
+4.  Create `/etc/ntopng/ntopng.conf` and match the contents to the example listed below. Replace `example.com` with your Linode’s domain or public IP address. If needed, replace `eth0` with your primary network interface. If you want to review available configuration parameters, run `man ntopng` from the terminal.
```
EdwardAngert: @Guaris @Darkstar90 : can one of you please confirm this change for me, please? I changed `[domain_or_ip]` and what I think was supposed to be a matching example (L103) at `--m="ipaddress"` with `example.com`

docs/networking/diagnostics/deploy-ntopng-on-debian-8.md
```diff
@@ -0,0 +1,180 @@
...
+
+    {: .file }
+    **/etc/ntopng/ntopng.conf**
+    : ~~~ conf
+      --user=ntopng
+      --interface=eth0
+      -w=3005
+      -m="example.com"
```
EdwardAngert: second instance here

docs/networking/diagnostics/deploy-ntopng-on-debian-8.md
Darkstar90: Hey Edward,

The "-m" option allows the user to specify local network addresses (i.e.  192.168.1.0/24). It wasn't included in my initial submission, so I don't know where it came from. Perhaps it found its way in by mistake. Here's what I think you're looking for. You'll want to place "example.com" in the "-w" option on line 102 like this: 

-w=example.com:3005

However, when I wrote this guide, I wanted to make the reader aware that if they did not own a domain, they could also substitute "example.com" with their Linode's public IP address. That was what [domain_or_ip] was supposed to signify. I hope that answers your question. If you have any other questions, just comment here and I'll be notified via email about it.

EdwardAngert: thanks for the clarification, @Darkstar90!
When I got a hold of the guide, there was a difference in example IP style, which is why I dug into the tech a little.
I also checked with @Guaris about where the option came from and
I'm going to update that step with an [example ip](https://linode.com/docs/linode-writers-formatting-guide/#example-ip-addresses) instead of `example.com` which will hopefully clear things up a little.

We're going to test things out and keep this going.
Thanks for your work on this!

-Edward

PR #737, opened 2017-07-13
Fix broken link on README

alexfornuto:
Based on the relative path, I assume you wanted this to point to the formatted version on the site, not the version on GitHub. As of now, it doesn't go anywhere.


EdwardAngert:
thanks, @alexfornuto  - you're right about the intention behind those relatives. I forgot that we were outside of the formatted land. Should be all fixed and [live on the site now](https://github.com/linode/docs)

PR #750, opened 2017-07-19
Update install-wordpress-on-ubuntu-16-04.md

jlvollmer:
Changing php.ini in the directory 'cli' does not work.  You must change php.ini in '/etc/php/7.0/apache2/php.ini' to increase upload size.


Guaris:
Thanks so much !

PR #756, opened 2017-07-20
Update obtain-a-commercially-signed-ssl-certificate-on-debian-and-ubu…

jerrykrinock:
…ntu.md

A bonehead like me could try `sudo apt-get update && apt-get upgrade`, and then spend some time wondering why this gave a 13: Permission denied error on my Ubuntu 16.04 LTS Linode.  I think the problem is that I needed to put another sudo after the &&.  It worked later like that.  But running the commands separately is probably a better idea for the caliber of people who use these guides :)


GuessWhoSamFoo:
Thanks for your input!

PR #774, opened 2017-07-25
Bring backups info into own paragraph and clarify

TheJosh:
It's really important that people know this.


Guaris:
Thanks for pointing this out. I've issued the fix.

PR #775, opened 2017-07-26
Fixed broken htop link in manage-processes-with-killall-and-kill.md

JoshuaRLi:
See [here](https://www.linode.com/docs/tools-reference/tools/manage-processes-with-killall-and-kill) under section `Finding Running Processes`. The `htop` docs link is broken and leaks ugly text, so I've fixed this.

Thanks for the docs, they're a great resource! :tada:


Guaris:
Thanks! It's a pleasure to get feedback from our community members!

PR #792, opened 2017-07-31
Update and rename deploy-a-production-ready-cassandra-node-cluster.md…

kcraig34:
… to deploy-a-production-ready-cassandra-node-cluster-on-ubuntu-and-centos.md

Copy edited. Needs to be reviewed before PR merged.


GuessWhoSamFoo:
Under `Generate SSL Files`, the configuration file sample which is supposed to have elements in bold has some issues rendering `**` inside the except. Might need to find workaround

PR #807, opened 2017-08-05
Add instructions for enabling Longview agent

rvl:
NixOS has a service for integration with Linode Longview.


GuessWhoSamFoo:
Thanks @rvl

PR #810, opened 2017-08-07
Edward angert jw dockerfiles

Guaris:


docs/applications/containers/how-to-use-dockerfiles.md
```diff
@@ -63,17 +63,18 @@ Note that in the example below, multiple packages are installed on separate line
         mkdir ~/mydockerbuild && cd ~/mydockerbuild
         touch Dockerfile
 
-2.  Open `Dockerfile` using a text editor and enter the following example to create a Dockerfile that installs `build=essential`, `curl`, and `make` onto a Ubuntu image:
+2.  Open `Dockerfile` using a text editor and enter the following example to create a Dockerfile that installs `build-essential`, `curl`, and `make` onto a Ubuntu image:
 
     {: .file }
     Dockerfile
     :   ~~~ docker
+        FROM ubuntu
```
EdwardAngert: @Guaris: I'm leaning towards leaving this as `ubuntu:latest` to keep it looking for only the latest LTS. What you think?

EdwardAngert: followup: according to https://www.fromlatest.io/#/ "base images shouldn't use the `latest` tag." So I might be wrong about this

Guaris: @EdwardAngert `latest` doesn't mean what it sounds like it means. So you are right about being wrong :)

PR #816, opened 2017-08-08
[UPDATE] Removed Amanda from Linode docs author

GuessWhoSamFoo:
Used `grep` recursively then `sed` to replace all instances of "Jane Doe"

`grep -ril "Jane Doe" ~/docsmith/source/docs | xargs sed -i "" 's/Jane Doe/Linode/g'`


GuessWhoSamFoo:
[Take out Amanda Folson as a writer](https://trello.com/c/aYvGMTq5/344-take-out-amanda-folson-as-a-writer)

PR #821, opened 2017-08-10
Use GiB like Linode Manager does

eatonphil:
In the screenshots included in the guide we write GiB but the docs say GB. This inconsistency is confusing (and inaccurate).


eatonphil:
Thanks!

PR #845, opened 2017-08-16
Update and rename mosh.md to Install-a-Mosh-Server-as-SSH-Alternative…

kcraig34:
…-on-Linux.md

Title, meta-description and other copy edits made. Please review alias and URL for accuracy.


GuessWhoSamFoo:
https://www.uottawa.ca/uoweb/en/user-guide/page-path-best-practices

The markdown file name becomes the URL path. Renaming to lower case per best practices.

kcraig34:
My bad, Sam. I was under the impression that the URL names automatically get switched to lower case by default. Guess not?
Will change upper to lower case in the future.
Thanks for proof of your argument.

PR #861, opened 2017-08-17
fixes markdown syntax for code block in note

brewn:
yep

also not sure why we're using curl in one and wget in the other but 🤷‍♂️


brewn:
also lol who wrote this?

> While the Linode Platform provides minimalist distribution templates and kernels complied to support the Linode's infrastructure explicitly, it is possible to deploy custom distributions and kernels within the context of the Linode Platform. These documents explore the use PV-Grub, for booting into a kernel of your choosing, and the procedure for creating and uploading custom distributions to run in your Linode instance."

It sounds like a robot picking words at random.

GuessWhoSamFoo:
Not really sure, it's definitely pre-2014 as there is no commit history. I do agree it is in need of revision.

PR #869, opened 2017-08-21
copy edits to half-life 2 deathmatch guide

EdwardAngert:


kcraig34:
@EdwardAngert -- Ditto. Pull SEO copy edit card into Ready and do - or assign to me. Then, pull image card into Ready and assign to Rob. Thanks.

PR #901, opened 2017-08-29
[NEW] How to install Mail-in-a-Box

zerocreativity:

PR #902, opened 2017-08-29
[NEW] How to install PrestaShop on Ubuntu 16.04

zerocreativity:


zerocreativity:
This depends on https://github.com/linode/docs/pull/901 since I've included a link to the Mail-in-a-Box tutorial as a way for the user to run his own email server. But the link can easily be removed for now if it creates problems/tutorials can't be published in order.

PR #912, opened 2017-09-01
Update NFS Doc to work with Debian 8

hjpx:
After troubleshooting this with a customer I wanted to share what needed to be done in this case to help others who might be in the same boat.

Following the Debian 7 guide, a user is left on the client with the mount command hanging indefinitely.


Guaris:
Awesome! Thanks a ton !

GuessWhoSamFoo:
@Guaris The filename needs to be updated as well.

PR #914, opened 2017-09-04
Setting-up tinc

damasosanoja:
Hi Linode team,

This is the article draft showcasing three different use-case scenarios. The first is explained in detail, the second is more an extension to remark tinc expandability and the third is a summary of everything learned.

Looking forward to hearing your suggestions,

Best regards.


Jskobos:
Hello @damasosanoja,

Thank you for the guide, it was well-written and interesting. Could I ask you to provide (as a comment here is fine, no need to commit) a sentence or two to describe when or why a user might want to try each of the three use cases, as well as what the benefits might be? e.g. the webserver-appserver-dbserver setup is presumably for someone deploying a web app, but this isn't spelled out. This way the guide will be useful for a wider range of readers.

Thanks,
Jared

damasosanoja:
Hello @Jskobos 

Thanks for your request, you are totally right about including an explanation about when/why use each topology and its benefits =)

**1.- (Two Nodes example):** A typical use case is a web-based invoicing software where database should be on a separate server (for security and disaster recovery)  and need to communicate sensitive data to the application server through Internet.
**2.- (Three Node example):** This topology can be considered as a variant of the first example where a high traffic application requires separate instances for better scalation, resource optimization and faster response (asynchronous processing). 
**3.- (Four-Node example):** This setup illustrates a use case where regional branches need to interact with an application running in the Head Quarters, e.g. a centralized inventory management software.

_I tried to be as brief as possible, please feel free to require more elaboration if needed._

Cordially,
DS

Jskobos:
Thanks @damasosanoja, these were exactly what we were looking for. One other thing: since the guide was on the long side, I've simplified a bit and taken out the CentOS instructions so that all of the servers use Ubuntu 16. Please provide a new version of the corresponding 'cheat sheets' images with the distribution row removed.

damasosanoja:
Hi @Jskobos updated cheat sheets committed, let me know if you need any other change ;)

Issue #907, opened 2017-08-30
Installation of MySQL on Centos 7 - Include wget


GuessWhoSamFoo:
https://github.com/linode/docs/blob/master/docs/databases/mysql/how-to-install-mysql-on-centos-7.md

Moved over from Jira, `wget` does not come with default installation of Centos 7. Add `yum install wget` for completeness.

GuessWhoSamFoo:
![](https://github.trello.services/images/mini-trello-icon.png) [Converting Jira tickets to Github issues (3PP)](https://trello.com/c/eUF6hRfF/399-converting-jira-tickets-to-github-issues-3pp)

Jskobos:
Issue resolved, #920

PR #919, opened 2017-09-05
[UPDATE] Replace/Add Pictures for August Freelance Graphics

Jskobos:
@EdwardAngert I moved the images to /assets and added links for the first ~5 images.

PR #920, opened 2017-09-05
[UPDATE] Wget mysql update

Jskobos:
#907 added wget to the Before You Begin section.


GuessWhoSamFoo:
Separating commit that followed from other PR: https://github.com/linode/docs/pull/919

PR #945, opened 2017-09-12
Sudo access for CentOS 6 Users

coocheenin:
Caution for CentOS Users about sudo access.


cwlinode:
Thank you for the addition.

PR #947, opened 2017-09-12
Added Note In Creating Folder public_html and logs

lynnellemmanuel:
I've added a note not to put space after comma in the command of making the folders public_html and logs. Beginners like me may accidentally put space after the comma, which will cause an error in reloading Apache. With the note, reader are made aware before they bump into an error that would discourage them from continuing the configuration.


GuessWhoSamFoo:
That sounds reasonable - thanks for the input!

lynnellemmanuel:
Thank you. Truly appreciated.

PR #958, opened 2017-09-13
Adds directory indices to vhost snippet.

nonprofitable:
I've followed this guide a few times, and always end up with errors because Apache shows me a directory listing instead of index.php or index.html. I thought I'd contribute this upstream instead of keeping it in my Linode setup notes.


GuessWhoSamFoo:
The section is about setting up virtualhost rather than an index page. Copying `/var/www/html/index.html` into `/var/www/html/example.com/public_html/` will show the default index page. The DirectoryIndex seems tangential to the point. Feel free to reopen if you think otherwise.

PR #957, opened 2017-09-13
[NEW] clamav quick answer from hphillips

EdwardAngert:
this one's a new guide from our own supportian @hjpx

PR #959, opened 2017-09-14
Update scanning-your-linode-for-malware.md

kcraig34:
Review complete.
We can use this guide, so let's put it on a Trello card and commence with the editing procedure on Sept. BurnDown Board.
If it (for some reason) gets drawn out, move it to October iteration (Content Priority and Ideas board).


GuessWhoSamFoo:
XR #957

PR #960, opened 2017-09-14
[New] Guide for setting up a Streisand gateway

Jskobos:


Jskobos:
@kcraig34 I noticed that the guide hadn't been added to the index, so I put it in. If all looks clear I'll merge it into master.

kcraig34:
If there's not problem, @Jskobos, go ahead and merge.

PR #966, opened 2017-09-14
Caddy Guide – Changes Requested

Jskobos:
Hello @coocheenin,

Thanks for the guide–it works well, and setting up Caddy as a service is a cool idea. Please edit the guide so that it works without a custom script. We prefer to avoid hosting scripts for our guides, as it makes the guides harder to maintain. In addition, it's much harder to install Daemonize on CentOS7, and it would be nice if the guide would work for that system as well. Perhaps you could try this [approach](https://github.com/mholt/caddy/tree/master/dist/init/linux-systemd)?

Jared


coocheenin:
Hello @Jskobos !
Thanks for review.

1. I made some changes. The external link to init-script listing **was removed** (the inline code box used instead). The caution box for CentOS 7 user's **was added**.
The custom init-script and Daemonize are required. Without these two things it's impossible to manage services (run, stop, etc.) confidently from regular (non-root) user on CentOS 6 and other init.d distros.

2. Don't think I'm slacking to write multipurpose guide! But I think it's wrong way to complicate this tutorial. CentOS 6 and 7 are equipped with totally different service management system: System-V and System.d. Commands `service` and `chkconfig` were replaced with `systemctl`. Also logging management was changed (`journalctl` command). The Caddy's init-scripts are very different from each other.

This all will require to write separate blocks for these cases. I don't think it's what we need for Linode Docs Simplicity. What do you think about, @Jskobos ?

coocheenin:
Oops, I have noticed your new edition. I make commits to my initial branch (PR 951). It's ok?

By the way. Your preface contains link to [Securing Your Server](/docs/security/securing-your-server) article. It's useless for CentOS 6 users until PR 45 will be not merged. For Newbies it's very important.

PS [If guide will be published]: Could you please edit contributor's data? Name And Lastname without abbreviations and Link to Github account. Thanks in advance!

Jskobos:
Hello @coocheenin,

After discussion here we've decided to make this a guide for installing Caddy on CentOS 7. This way the guide will be more up to date, have a longer usable lifespan, and be simpler for the users. I'm going to submit the guide for a copy edit today, and we will make sure to include your full name.

Thanks,
Jared

coocheenin:
Sad about CentOS 6 version, but OK.

coocheenin:
Hello @Jskobos 

A few corrections here.

It's very important **to extend capabilities** for www-data user, because Caddy will listen ports < 1024. Let's add somewhere near permission's setup: `sudo setcap 'cap_net_bind_service=+ep' /usr/local/bin/caddy`
The custom init-script contains this command, but original - not.

And after words _Create an unprivileged user, which will be used for running your Caddy server._ (string 55) it seems two commands are missing:
`sudo groupadd www-data` and `sudo useradd www-data -d /var/www -g www-data -s /sbin/nologin`.
This user does not exists by default (and group too), right? In CentOS 6 surely not.

Jskobos:
Hello @coocheenin,

Thanks for your input. I've added the `setcap` command along with some of the other permissions checks from the official docs. 

Jared

coocheenin:
Hello @Jskobos 

Linode Docs Rocks! What's next?

cwlinode:
@coocheenin Your guide is being edited internally and should be published soon. Closing these outstanding PRs.

PR #980, opened 2017-09-19
Visualizing Apache Webserver Logs in the Elastic Stack on Debian 8

tylerjl:
Howdy, here's the September guide for Visualizing Apache Webserver Logs in the Elastic Stack on Debian 8. This guide takes up from the [Apache Web Server on Debian 8 (Jessie)](https://www.linode.com/docs/web-servers/apache/apache-web-server-debian-8) guide and ends with Elasticsearch, Logstash, and Kibana configured and an initial tutorial regarding how to set up a visualization. I considered adding another visualization tutorial, but the guide is already fairly information-heavy with the three pieces of the stack included, so if it's sufficient, then great, otherwise I can add another section and/or just save that for another dedicated guide on creating visualizations in Kibana if that's more convenient.

Thanks!


GuessWhoSamFoo:
Closed in favor of #993

PR #1009, opened 2017-09-28
[NEW] ELK stack security visualization – A. Lescher [Changes Requested]

Jskobos:
Hi @Darkstar90,

I'm going through this guide with Ubuntu 16 and Debian 9, and there are some things not working (Java and ElasticSearch installation.) Please edit to focus on one distro where we can verify all of the commands (preferably the most up-to-date version of Debian, Ubuntu, or CentOS). That way we can also simplify the guide by removing things like the Python section–since with any up-to-date distro the Python version will already be 2.7. 

Thanks,
Jared


Jskobos:
Hi @Darkstar90, any progress on this? 

Thanks,
Jared

Darkstar90:
Hey Jared,

I have the guide narrowed down to just CentOS 7 instructions as requested, and removed the Python section as it was no longer relevant. I think you'll need to open the PR I created to view the changes, as they are not showing up here.

As for the errors you ran into, I'll need a bit more information to go on. If they were CentOS 7 related, what errors were you encountering when installing Java/ElasticSearch?

Thanks,
Andrew

Jskobos:
Hi Andrew, the problems were with Ubuntu/Debian, so those sections work fine now. I've gotten Wazuh/Kibana to work as written. 

I'm having trouble getting the subdomain to redirect, but not sure if that's because we're using a Linode without a FQDN (we've been trying with the IP address and with kibana.XXX.members.linode.com). How did you have this set up?

Darkstar90:
Hey Sam,

Sorry for the delayed response - for some reason I didn't see the second paragraph in your reply. I can definitely look into this tomorrow (Tuesday) and try to determine the problem with your setup. You appear to have the URL setup correctly from what I can tell, and I don't think not having an FQDN is the problem.

However, with the limited info I have, I would assume it's just not routing correctly to the Linode you're working on. If you could, I'd like you to check if your Linode "owns" the linode.com domain name. In other words, is linode.com associated with your Linode's IP address (this is setup in the DNS page in Linode Manager). 

I would also like to mention that we could just remove this section, as it's not necessary to complete the guide. The main advantage it offers is access via a clean URL (instead of an IP address and port #) and security with SSL encryption. If you think the guide should just cover installation, I would recommend we take this part out.

Let me know what you'd like to do.

PR #1030, opened 2017-10-05
Update NixOS installation guide for NixOS 17.09

nixy:
NixOS 17.09 nicknamed "Hummingbird" was released so I went through this guide to ensure that it still worked.

Everything in the guide still worked, but running through it again I found that downloading the installer could use a touch up. I also realized there was a better way to configure grub to get only the configuration file.

These are small changes and the guide is still good, but this will make the user experience a little smoother.


kcraig34:
Thanks, @nixy. Update commits will be merged in next publishing.

PR #1040, opened 2017-10-11
[NEW] Deep learning API guide

Jskobos:
Ready for tech edit.


GuessWhoSamFoo:
Some things that might come up in the future:

Handling of async events by web server: https://github.com/fchollet/keras/issues/2397

Pre-trained models optimizer weight exception: https://github.com/fchollet/keras/issues/4044

Lazy loading and WSGIImportScript: https://modwsgi.readthedocs.io/en/develop/user-guides/frequently-asked-questions.html

docs/applications/big-data/how-to-move-machine-learning-model-to-production.md
```diff
@@ -0,0 +1,306 @@
...
+
+        model.save('my_model.h5')
+
+        ~~~
+
+3.  Run the script:
+
+        python ./mnist_model.py
+
+    There may be a warning message as shown below from a pip or conda installation which means installing from source could offer superior performance.
```
EdwardAngert: @Jskobos this says there might be a message about installing from source over the method here, but it doesn't say to or not to. Should we suggest that the reader ignore the message?

Jskobos: @EdwardAngert Sorry, just saw this. That comment was from Sam, yes the warning can be ignored.

Jskobos: Idea was to clarify that it's possible to speed things up if the reader is up to it.

PR #1050, opened 2017-10-16
How to use ZFS on Ubuntu 16.04

zerocreativity:


Guaris:
@zerocreativity Hello, I am one of the editors at Linode. I went through your guide, and everything is good. I had a question about the final section. 

```

    By default, the ARC will use at most 75% of total memory. While this makes sense on a server with multiple purposes (e.g. also running database software), on a storage server it's a loss of approximately 20% (or more)
RAM that could be put to good use. So follow these instructions only if you're not using your Linode to run other (memory hungry) applications.

The numbers in the next command represent bytes. 1GB is calculated by multiplying 2 to the power of 30 but you can use values such as 7000000000 to reserve approximately 7GB. As a rule of thumb, set zfs_arc_max to total
memory available on your Linode minus 1GB.  zfs_arc_min can be set to 50% of total RAM (if more than 2GB is available, at least). Copy the lines in a text editor, change the numbers and then paste all lines at the same time:

        cat <<EOF > /etc/modprobe.d/zfs.conf
        options zfs zfs_arc_min=3221225472
        options zfs zfs_arc_max=6442450944
        EOF

```

Are you suggesting that I open a text editor and edit `zfs.conf` ? Like `vi /etc/modprobe.d/zfs.conf` and add the four lines? If not, could you provide clarity? Also, after the lines are added, the guide sort-of ends very abruptly. Maybe I am misunderstanding, but shouldn't we restart a service to pick up the new change?  Let me know. 

angel

zerocreativity:
Good catch, yes the ZFS module has to be reloaded. But rather than unmounting, stopping services and finally modprobe to remove and reload the module, of course, it's easier to just reboot.

Here's an updated text:

```
Copy the lines below in a text editor (e.g. Notepad) so you can adjust the two numerical values before pasting the command on your server.

        cat <<EOF > /etc/modprobe.d/zfs.conf
        options zfs zfs_arc_min=3221225472
        options zfs zfs_arc_max=6442450944
        EOF

Now reboot the machine for these changes to take effect.
```

Tell me if I need to commit this. Also, If it's still unclear, then I can just change this to the nano, paste, edit, save routine.

Guaris:
Can you change this to the nano, paste, edit, save routine? I'll add it into the guide, no need to commit. Thanks a lot !

zerocreativity:
```
By default, the ARC will use at most 75% of total memory. While this makes sense on a server with multiple purposes (e.g. also running database software), on a storage server it's a loss of approximately 20% (or more) RAM that could be put to good use. So follow these instructions only if you're not using your Linode to run other (memory hungry) applications and you have at least 4GB of RAM.

The numbers in the file you will create (`zfs.conf`) represent bytes. 1GB is calculated by multiplying 2 to the power of 30 but you should use values such as 7000000000 to reserve approximately 7GB. As a rule of thumb, set **zfs_arc_max** to total memory available on your Linode minus 1GB.  **zfs_arc_min** can be set to 50% of total RAM. Example: if you have 16GB available, zfs_arc_max would be 15000000000 and zfs_arc_min would be 8000000000.

Run the following command:

        nano /etc/modprobe.d/zfs.conf

Paste these lines into the editor and then adjust the two numerical values:

        options zfs zfs_arc_min=3221225472
        options zfs zfs_arc_max=6442450944

Now reboot the machine for these changes to take effect.
```

Shifted even more things around to avoid any possible confusions users may have. Hope this is good @Guaris.

Guaris:
Thanks a million @zerocreativity !

Guaris:
#1069

PR #1061, opened 2017-10-17
updated dd command to show progress

sopsmattw:
I updated the dd command to include pv, which provides a progress indicator


GuessWhoSamFoo:
@Guaris It doesn't really take that long to complete to merit a progress bar IMO. Thoughts?

Guaris:
@GuessWhoSamFoo Agreed.

PR #1103, opened 2017-10-28
Update how-to-mount-nfs-shares-on-debian-8.md

gacanepa:
- Removed step #3 in the NFS Server Setup section since portmap is a virtual package that's satisfied with the presence of rpcbind, which is installed by default in Debian 9 Stretch.
- Added code syntax (conf) for file excerpts
- Replaced network preffix /32 with /17, as indicated in Private IPv4 under https://www.linode.com/docs/networking/linux-static-ip-configuration#static-network-configuration
- Replaced service command with systemctl equivalent
- Indicated how to mount NFS shares manually using the mount command with the -o switch
- Added noexec in the list of mount options (/etc/fstab entry)
- Added root_squash in the list of server options, and explained the implication of using it and no_root_squash
- Added Examples section
- Minor formatting changes


Jskobos:
Closed in favor of #1107

PR #1107, opened 2017-10-30
[UPDATE] How to Mount NFS Shares on Debian 9

Jskobos:


Jskobos:
@EdwardAngert or copy editor: I made a new file for the update, since it's for Debian 9 now. We have to decide whether to delete or deprecate the older guide.

GuessWhoSamFoo:
Debian 8 EOL is 2020. I don't think we should deprecate although it appears redundant in many areas.

kcraig34:
@GuessWhoSamFoo -- It won't kill us to run with both old and new versions for a couple of years.
So we can keep 'em and revisit in 2018?
Thanks for the advice.

cwlinode:
Closing and picking up with https://github.com/linode/docs/pull/1144

PR #1109, opened 2017-10-31
Update and rename using-haproxy-for-load-balancing.md to how-to-use-h…

kcraig34:
…aproxy-for-load-balancing.md

SEO and 2nd CE completed. Please review, revise as necessary, merge the PR and publish the doc.
Thanks.


GuessWhoSamFoo:
@EdwardAngert Don't forget an alias since there is a change to filename

PR #1131, opened 2017-11-03
CAA records quick answer

cwlinode:


kcraig34:
@cwlinode -- Solid effort on this. Looks like it also needs a copy/SEO edit? 
If so, can you copy the card its now on and put a CE/SEO edit card in Ready column? Thanks.

PR #1134, opened 2017-11-04
Add "nameserver" options to config file

ppggff:
The "nameserver" option is used inside the ss-server(shadowsocks server) to resolve hostname requested from client,
skipping it may result very slow response from the server.

For example, I got a big improve after set this option: from 3xxxms to 5xxms.


Guaris:
@ppggff Approved! Thank you for your submission.

PR #1145, opened 2017-11-07
Replace dashes

Jskobos:
@GuessWhoSamFoo Here are the updated docs (with dashes removed). There is a list of altered files in scripts/results.txt.


GuessWhoSamFoo:
Some of the config files and possibly tables are being affected. I will manually remove those changes later.

PR #1152, opened 2017-11-08
Fix typo in "mtr" TCP mode example.

martey:
The `-c` parameter is missing from these examples. `-rw50` is not valid mtr syntax.


GuessWhoSamFoo:
Looks like it was wrong from the beginning: https://github.com/mobolic/linode-docs/commit/8a103739fa270e400de9834281d441a01133d226#diff-e137dd533837bfe732aa7c65eb6c917aR374

Thanks @martey

PR #1166, opened 2017-11-13
add get temporary password command and example display

RainC:
mysql_secure_installation requires  root password ( CentOS 7 tested ) 

So I add get temporary root password before "mysql_secure_installation" 

Signed-off-by: DongGyu Kim <me@rainclab.net>


GuessWhoSamFoo:
By default, an installation of mysql-server has a root user without password. Not sure why one would want to create a temporary password just to change it again during the next step. Feel free to reopen if you feel strongly otherwise.

PR #1167, opened 2017-11-13
Archetype

Jskobos:
New archetype for generating docs.


GuessWhoSamFoo:
@Jskobos This change should be made in `linode/hugo`

GuessWhoSamFoo:
@Jskobos Let's make the target branch master since `hugo_format` was already merged without this included

Jskobos:
@GuessWhoSamFoo I think we can delete this PR, the archetype already ended up in master somehow.

PR #1190, opened 2017-11-14
Fix arping command

creativetags:
arping command was wrong according to the man page entry. I've changed it to what worked for me.


GuessWhoSamFoo:
XR (https://jira.linode.com/browse/DOCS-52)

Thanks @creativetags

PR #1191, opened 2017-11-14
made a quick find this guide on github page

EdwardAngert:
in /docs, but might need to be moved because Hugo didn't seem to care for it locally. Will tinker with pre-flight


Jskobos:
Closed in favor of #1192.

PR #1196, opened 2017-11-15
Update and rename visualize-server-security-with-the-elk-stack.md to …

kcraig34:
…visualize-data-security-with-elastic-stack-and-wazuh-on-centos-7.md

Changes made per Elasticsearch brand preferences and to better capitalize on SEO for headers.
Please review carefully new alias in metadata.

PR #1206, opened 2017-11-16
Visualize logs with elastic and wazuh

EdwardAngert:
still not totally positive where this went so wrong, but this should right the wrongs

incorporates some of @kcraig34's suggestions from https://github.com/linode/docs/pull/1196 in their own commit (since he had been working on an older version, not all changes will show in the commit (somethings, like `ElasticSearch` > `Elasticsearch`, were already fixed))

deletes the duplicate file + updates the graphic


kcraig34:
Excellent, @EdwardAngert!
Another hurdle leapt!
Thanks.

PR #1210, opened 2017-11-17
change Caddy instructions to use EPEL package

carlwgeorge:
Caddy is now available via the EPEL repository.  Using that package in these instructions avoids having to perform many of these setup steps manually.  This PR also removes some out of scope instructions that would have needed to be updated to work with using the Caddy package.


GuessWhoSamFoo:
Thanks for the update!

PR #1224, opened 2017-11-21
[UPDATE] Ruby on Rails with Passenger on Debian 9

Jskobos:


kcraig34:
I'm wondering if we should apply an SEO edit to these guides that get a updated...this guide could use an og descptn, "What is RoR?" H2 header, improved meta descrptn, etc.
Should we consider a 2-step process, update then SEO edit, @Jskobos @EdwardAngert

PR #1225, opened 2017-11-22
Update how-to-set-up-tinc-peer-to-peer-vpn.md

reed-jones:
Updated to Tinc v1.0.33 and added https to download link


GuessWhoSamFoo:
Thanks broski

PR #1240, opened 2017-11-28
[New] Monitor an nginx Web Server Using the Elastic Stack on Centos 7

tylerjl:
Here's the guide for November! 📝 

Notes:
- This borrows heavily from the previous Elastic Stack on Debian guide, so I took the opportunity to expand the scope from just webserver logs to machine metrics as well.
- The section regarding Kibana access presented a tricky question re: security since Kibana does _not_ provide user authentication as part of the vanilla OSS distribution. I addressed this by briefly explaining a locally-forwarded ssh tunnel, which should work and doesn't expose any publicly-reachable services to the general internet.
- This uses v6 of the stack which was released a week ago (to much fanfare), so a fair bit of this content is new/not covered elsewhere.

Let me know if there's anything else required on my part. _waves to @Guaris_


GuessWhoSamFoo:
Closed in favor of new PR

PR #1243, opened 2017-11-28
Deprecate SVN guide

Guaris:


GuessWhoSamFoo:
Still don't think this should be deprecated...

kcraig34:
@GuessWhoSamFoo -- Discuss with team there in HField, or let's discuss at meeting on Thursday.

PR #1261, opened 2017-12-01
Shortguides

Jskobos:
Rebuild theme and add new folder `shortguides`, with example shortguide for installing Python 3 with Miniconda.


kcraig34:
Do we have to communicate with writer about level of contribution in this situation, @Jskobos?
Please advise as to how much remains is original and how much we wrote?
Thanks.

PR #1262, opened 2017-12-01
Add "Installing a Minecraft Server inside a FreeBSD Container"

RainbowHackerHorse:

PR #1292, opened 2017-12-11
[NEW] Install Minecraft inside FreeBSD Container

GuessWhoSamFoo:
XR #1262 

TODO:

- [x] Break apart numbering
- [ ] Change to meet writing style of current docs

### DIFF
--- a/docs/game-servers/install-minecraft-in-a-freebsd-jail.md
+++ b/docs/game-servers/install-minecraft-in-a-freebsd-jail.md
@@ -5,7 +5,7 @@ author:
 description: 'FreeBSD''s advanced security features can be used to safely provide a secured container for hosting game servers, like Minecraft.'
 keywords: 'freebsd,bsd,minecraft'
 license: '[CC BY-ND 4.0](https://creativecommons.org/licenses/by-nd/4.0)'
-modified: Friday, July, 08, 2016
+modified: Tuesday, July, 25, 2017
 modified_by:
   name: Rainbow
 published: 'Monday, March, 14, 2016'
--- a/docs/game-servers/install-minecraft-in-a-freebsd-jail.md
+++ b/docs/game-servers/install-minecraft-in-a-freebsd-jail.md
@@ -2,13 +2,13 @@
 author:
   name: Rainbow
   email: rainbow@hacker.horse
-description: 'FreeBSD''s advanced security features can be used to safely provide a secured container for hosting game servers, like Minecraft.'
-keywords: 'freebsd,bsd,minecraft'
+description: "FreeBSD's advanced security features can be used to safely provide a secured container for hosting game servers, like Minecraft."
+keywords: ["freebsd", "bsd", "minecraft"]
 license: '[CC BY-ND 4.0](https://creativecommons.org/licenses/by-nd/4.0)'
-modified: Tuesday, July, 25, 2017
+modified: 2017-07-25
 modified_by:
   name: Rainbow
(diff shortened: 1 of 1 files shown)


GuessWhoSamFoo:
@RainbowHackerHorse So it looks like the current version of `iocage` has a broken installation on FreeBSD 11.1 (See https://github.com/iocage/iocage/issues/421). Did you end up compiling iocage back when this was written?

RainbowHackerHorse:
@GuessWhoSamFoo I wrote this back when `iocage-legacy` was still the version of `iocage` being utilized.
It looks like everything should be back to normal with `py-iocage` now, but tbh, I think legacy was deprecated early anyway >.> `py-iocage` isn't production ready, as is shown here.
One can use the package from the following link:
[https://github.com/RainbowHackerHorse/Stuffnthings/raw/master/legacypkgs/iocage-1.7.6.txz](url)

To install it directly, one can run the following commands:
`pkg install -y ca_root_nss`
`pkg add https://github.com/RainbowHackerHorse/Stuffnthings/raw/master/legacypkgs/iocage-1.7.6.txz`

RainbowHackerHorse:
It actually looks like there's a maintained fork of iocage legacy called `iocell` and its in ports. I'm going to toy with it, and test this guide with it. If it works, I'll pull down the changes you've made, and update it to use `iocell`

GuessWhoSamFoo:
@RainbowHackerHorse Any updates on this?

RainbowHackerHorse:
s/iocage/iocell and you're good.
It's drop-in compatible.
Haven't had a chance to make the edit myself but it's just a quick find/replace-all

RainbowHackerHorse:
Yeah, since you slurped up the branch, it's a bit harder for me to make the edit now :P

GuessWhoSamFoo:
If just substituting, should be fine - I'll take care of it

RainbowHackerHorse:
@GuessWhoSamFoo Aye, it's just substituting the command name itself. Usage is exactly the same :)

GuessWhoSamFoo:
@RainbowHackerHorse Sorry this took a while. Priorities were shifted around.

Anyhow, I'm running into two issues:

1) `iocell` loses connectivity to the outside world. Entering the jail's console will fail `pkg` and pinging anything. From a Google search, it seems to be a common issue with iocage as well. Any leads on this one?

2) I proceed with the rest without the jail. After running `make all-depends-list | cut -c 12- | xargs pkg install -y`, the message below showed up:

screen shot 2018-02-07 at 4 29 01 pm

Is installation of JDK 8 through ports the recommended way?

RainbowHackerHorse:
No worries!
If connectivity fails, you just need to stop the jail ( `iocage stop $jailname`), restart pf if you haven't (`service pf restart`), and start the jail (`iocage start $jailname`)
Once the pf rules have been added, it is imperative that pf is restarted.
That being said, ping is normally disabled from inside a jail, so that shouldn't be used as a metric for network connectivity. If pkg is failing, then try the steps I outlined :)

Installation via pkg is the recommended way for the JDK (which is what that command is doing), and this is a normal message.
We can update the guide to include that information, but I feel most people would be knowledgeable enough to follow the instructions they'd receive here.

RainbowHackerHorse:
Why was this closed? Has it been imported into the docs Library? AFAIK this was still active and pending.

Guaris:
@RainbowHackerHorse Hey! I closed it. The person assigned to work on this is no longer with us and we don't have the bandwidth to tackle this guide.

PR #1299, opened 2017-12-13
Add show_on_rss_feed param to appropriate guides.

Jskobos:
As part of change to Hugo templates, guides with show_on_rss_feed:
false will not be tracked on the RSS feed. This flag was applied
to special guides including formatting guide, contribute topics
list, and Github redirect page.


kcraig34:
Good house-cleaning, @Jskobos. Thanks.

PR #1306, opened 2017-12-14
Update using-rdiff-backup-with-sshfs.md

imfatant:
I should have done this in the first place. Step 3 becomes unnecessary, since ssh-copy-id does exactly that..

- Imfatant


GuessWhoSamFoo:
Looks like #1249 needs to be re-visited. The documentation changed from local -> remote to remote -> mount.

PR #1321, opened 2017-12-15
[NEW] Celery and RabbitMQ Guide

Jskobos:
Tech edited, ready for copy edit.


docs/development/python/task-queue-celery-rabbitmq.md
```diff
@@ -0,0 +1,363 @@
...
+{{< /output >}}
+
+3.  Use the **inspect stats** command to get statistics about the workers. It gives lot of informations, like worker ressource usage under `rusage` key, or the total tasks completed under `total` key.
+
+        celery -A downloaderApp inspect stats
+
+
+## Monitor your Celery cluster with Flower
+
+Flower is a web-based monitoring tool that can be used instead of the `celery` command.
```
EdwardAngert: @Jskobos do we want to include a "why" here? Something like, "Use Flower because it smells nicer than celery"

otherwise, we're just suggesting people use this instead of that 1:1, without reason

Jskobos: It uses a web browser, so it's more colorful than the CLI. Could be worth adding that bit instead of the 'the celery command' clause.

Also the output tags won't format right quite yet but please leave them as is. I'm going to try to merge in the new output shortcode today or Mon morning.

EdwardAngert: [STET]

looked at this again, within context - disregard my comment. It's a UI and it's nice looking.

PR #1322, opened 2017-12-16
[New] Nginx virtual hosting guide

Darkstar90:
This guide details the process for setting up virtual hosting with the Nginx web-server.


cwlinode:
@Darkstar90 Thank you for your submission but we unfortunately will not be moving forward with this guide. We've been in the process of redoing our NGINX guides for some time now, and hosting multiple server blocks is only a small addition to the config which doesn't require an entirely new guide.

PR #1342, opened 2017-12-27
Use tilde expansion for installation guides

liamdemafelix:
Some users won't notice the path, since it isn't highlighted. Granted, users should read before copy-pasting but when we're catering to new people, that's one less question to deal with.


GuessWhoSamFoo:
There are probably more cases of this all over the library. For guides that are simply about installation, this is probably the better option.

PR #1347, opened 2017-12-28
Correct chmod command for setuid bit

ErikSwan:
The `chmod` command for setuid should be `u+s`, not `o+s`.


Jskobos:
Thanks for the fix.

PR #1354, opened 2017-12-30
How to Benchmark Linux Servers

zerocreativity:


GuessWhoSamFoo:
I fail to see how this is different than https://joshtronic.com/2017/07/23/five-dollar-showdown-linode-digitalocean-lightsail-vultr/ especially within the single core space.

kcraig34:
@GuessWhoSamFoo -- Except for the fact that this would be a Linode boilerplate in our library and not on a third-party blog, the 2 are quite similar. But then, Linode does periodic attempts to benchmark vs. competitors. So, having such a guide in our library isn't the worst thing possible.
Unless you really are adverse to publishing AA's "guide" - more of a why-to than a how-to - let's continue with editing and publishing - unless team consensus is otherwise.

GuessWhoSamFoo:
@zerocreativity Did you run this on a VM? I'm getting different numbers on a Linode.

zerocreativity:
@GuessWhoSamFoo I ran it on a Linode 1024 in the Frankfurt datacenter.

PR #1355, opened 2018-01-01
Elasticsearch Plugin Guide

tylerjl:
Per previous conversations, here's a guide outlining how to install and use several different Elasticsearch plugins. Tested on latest Ubuntu and CentOS 7.

I don't think there's anything content-wise missing here; the only step that isn't completely explicit is the blurb explaining how to interact with the Elasticsearch REST API, but I tried to offer as many options as possible (in-editor, command-line with `curl`, browser based with Kibana Console, etc.)

Cheers!


kcraig34:
Thanks, Tyler.
We're in receipt of PR and moving it through the editing process.
Welcome to 2018!

Guaris:
@tylerjl 👋

Guaris:
Closed in favor of #1368  !

PR #1370, opened 2018-01-08
Cygwin for windows instructions

Guaris:
Reference to Jira-Ticket 55


Guaris:
@cwlinode Heyo can u eye this over and smash it in

PR #1378, opened 2018-01-09
removing parenthetical in image debian 8 lamp

EdwardAngert:
@kcraig34


kcraig34:
@EdwardAngert -- Please also check that image visualizes...

PR #1380, opened 2018-01-11
disable smart fractions extension for blackfriday

stvnjacobs:
This extension, which is enabled by default, will render anything
resembling a fraction, as a fraction. I can't think of any instances
where we use math in our guides, and we mention 24/7 in at least one of
them. This will disable the automatic conversion.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -22,6 +22,9 @@ disableKinds = ["taxonomy", "taxonomyTerm"]
 
 rssLimit = 15
 
+[blackfriday]
+fractions = false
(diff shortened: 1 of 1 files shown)


stvnjacobs:
example: https://linode.com/docs/platform/linode-managed/

stvnjacobs:
it will still convert halves and quarters to their appropriate html entity.

hugo blackfriday docs: https://gohugo.io/getting-started/configuration/#blackfriday-options

PR #1381, opened 2018-01-11
Copy edits to Terraform guide

Jskobos:


GuessWhoSamFoo:
What was the driving factor on this edit? XR #1098

PR #1394, opened 2018-01-12
Add script for identifying old non-deprecated guides

Jskobos:


GuessWhoSamFoo:
@Jskobos `requirements.txt` needs to be updated

PR #1403, opened 2018-01-15
Alpine Linux tech fixes

keithieopia:
* 0f3cddb - Alpine's grsecurity kernel has changed `vmlinuz-grsec` and `initramfs-grsec` filenames to `vmlinuz-hardened` and `initramfs-hardened`. **As the guide is currently written, GRUB2 will fail to boot Alpine.**

* be60d11 - the default `/dev/sudoers` recommends using the `wheel` group. It's a built-in group in Alpine, so for the sake of simplicity, the guide can be updated to remove the extraneous step to setup a new group.


GuessWhoSamFoo:
@cwlinode

cwlinode:
@keithieopia Thanks for the update.

PR #1404, opened 2018-01-15
Find old guides

Jskobos:


GuessWhoSamFoo:
Thanks for updating the docstrings. Looks good! This probably should live outside of ci/ but it's fine where it is for now.

PR #1431, opened 2018-01-22
Install an Odoo 11 Stack on Ubuntu 16.04 Using Linode

damasosanoja:
Hi @Jskobos 

Given the date, it seems convenient to me not to delay this article any longer. I am sending the original version written past December, the only change I made was to remove the Terraform section since the guide seemed very long.
The idea of using Docker still seems good to me, I'll keep it for version 12 of Odoo :)

To all your team congratulations on the idea of being able to visualize the article locally with Hugo. It is simply fantastic and very useful.


damasosanoja:
Hi @Jskobos 

Stage branch was removed from the guide and whitespaces check is ok now.

Jskobos:
Closed in favor of #1440

PR #1461, opened 2018-01-30
[NEW] Split Files on Linux

Jskobos:


GuessWhoSamFoo:
This needs to include instructions for a Mac using `gpslit`

PR #1465, opened 2018-01-30
Split mysql backup doc and update

cwlinode:


cwlinode:
The 404's are ok. These two docs are new and reference each other.

PR #1488, opened 2018-02-03
Update how-to-use-yubikey-for-two-factor-ssh-authentication.md

cassidycodes:
The example snippet in step 5 was missing the `key` argument. I added it here and changed the formatting to match an example of the same snippet that appears later in the document.


Guaris:
@cassidycodes Hey thanks for catching this!

PR #1491, opened 2018-02-05
Manage a Docker Cluster with Kubernetes

damasosanoja:
Hi @Jskobos

This guide requires the user to complete the previous [Kubernetes guide](https://linode.com/docs/applications/containers/how-to-deploy-nginx-on-a-kubernetes-cluster/).  I had problems using Linode 2GB plans (insufficient CPU resources), had to use 4GB instead. I assume it's because of the newer Kubernetes 1.9 version.


Jskobos:
Thanks Damaso. We won't be starting the editing process on this one until at least next week, but we'll let you know once we've had a look.

damasosanoja:
Hi @Jskobos

Thanks for the update.

Guaris:
XR #1524

PR #1493, opened 2018-02-05
[NEW] How to Use LXD Containers on Linode

Jskobos:


GuessWhoSamFoo:
From a cursory glance, I don't think "machine container" is an actual term. Probably not a good idea to coin it here.

PR #1613, opened 2018-03-16
Netlify deploy test

GuessWhoSamFoo:
Test


GuessWhoSamFoo:
The BaseURL is still throwing off any JS or assets that uses /docs. To be fixed in an upcoming iteration.

PR #1619, opened 2018-03-18
minor change - corrected misspelling

JohnBlood:
Changed "hot" to "how" and deleted an unneeded comma


Jskobos:
Thanks for the catch @JohnBlood!

PR #1630, opened 2018-03-20
Update floating IP to failover IP

GuessWhoSamFoo:
Not really sure the best way to disambiguate these terms - or if we want to at all.

virtual ip vs floating ip vs failover ip

Thoughts? @Jonchun @EdwardAngert


Jonchun:
These are all terms that refer to the same thing. Within the Linode platform, they are called _failover IPs._ I don't think there is a need to try and disambiguate like this, and rather just note via something like `Note: Some other common terms for referencing a "failover IP" include "floating IP" and "virtual IP". `

PR #1636, opened 2018-03-23
[NEW] Use NGINX as a Reverse Proxy

Jskobos:


docs/development/nodejs/install-nodejs-ppa.md
```diff
@@ -2,14 +2,14 @@
...
+description: 'Install an up to date Node.js binary from NodeSource on Debian or Ubuntu.'
 keywords: []
 license: '[CC BY-ND 4.0](https://creativecommons.org/licenses/by-nd/4.0)'
 published: 2018-03-06
-modified: 2018-03-06
+modified: 2018-03-26
 modified_by:
   name: Linode
-title: 'Install Node.js using an Ubuntu ppa'
+title: 'Install Node.js from NodeSource on Debian or Ubuntu'
```
cwlinode: We should also rename this file if an alias doesn't mess with the shortcode.

EdwardAngert: @cwlinode I didn't include it in the copy edit, but it should be easy enough to find/replace instances of it if everyone's into it. Let me know and I'll update

cwlinode: If it doesn't break anything already in existence, then i vote renaming it to `install-nginx-ubuntu` or something similar to the centos version. It really just needs the PPA removed from the title.

Jskobos: This will break any guides that already refer to that shortguide. Pretty easy to find and fix those though.

EdwardAngert: renamed to `install-nodejs-nodesource.md`

PR #1645, opened 2018-03-27
Fix false negative case for mixed whitespace test

### LINKED ISSUE #1657: Remove mixed whitespace in documentation
#1645  Fixes the regex for detection of mixed whitespace. The remaining offending documentation should have tabs converted into four spaces as per the style guidelines.

```
./docs/web-servers/apache-tips-and-tricks/tuning-your-apache-server.md
./docs/uptime/monitoring/top-htop-iotop.md
./docs/tools-reference/tools/load-testing-with-siege.md
./docs/platform/disk-images/clone-your-linode.md
./docs/platform/linode-backup-service.md
./docs/networking/vpn/install-openvpn-access-server-on-linux.md
./docs/networking/dns/configure-your-linode-for-reverse-dns.md
```

GuessWhoSamFoo:
Closes #1657 

Correct missed cases prior to merging

### DIFF
--- a/ci/test_whitespace.py
+++ b/ci/test_whitespace.py
@@ -18,9 +18,11 @@ def test_whitespace(md_filepath):
 @file_io
 def test_tabs(md_filepath):
     has_tabs = False
+    indent_regex = regex.compile(r'([ \t]*)')
     for line_number, line in enumerate(md_filepath, 1):
-        has_tabs = '\t' in list(line)
-        if has_tabs:
-            print("Mixed spacing on line " + str(line_number))
+        for position, char in enumerate(indent_regex.match(line).group(1)):
+            has_tabs = char != ' '
+            if has_tabs:
+                print("Mixed spacing on line " + str(line_number))
     assert has_tabs == False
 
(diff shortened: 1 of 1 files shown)


GuessWhoSamFoo:
Resolved cases via #1659 and rebased.

ci/test_whitespace.py
```diff
@@ -18,9 +18,11 @@ def test_whitespace(md_filepath):
 @file_io
 def test_tabs(md_filepath):
     has_tabs = False
+    indent_regex = regex.compile(r'([ \t]*)')
     for line_number, line in enumerate(md_filepath, 1):
-        has_tabs = '\t' in list(line)
-        if has_tabs:
-            print("Mixed spacing on line " + str(line_number))
+        for position, char in enumerate(indent_regex.match(line).group(1)):
```
Jskobos: Why are you enumerating here if the position character isn't being used?

ci/test_whitespace.py
```diff
@@ -18,9 +18,11 @@ def test_whitespace(md_filepath):
...
+            has_tabs = char != ' '
+            if has_tabs:
+                print("Mixed spacing on line " + str(line_number))
```
Jskobos: If there are multiple tabs in a single line, would this print identical output multiple times?

ci/test_whitespace.py
GuessWhoSamFoo: Should print `\t` positions now

ci/test_whitespace.py
GuessWhoSamFoo: This should not be identical anymore with the inclusion of position in the output now

PR #1651, opened 2018-03-28
[NEW] Manage Python Packages and Virtual Environments with Pipenv

Jskobos:


docs/development/python/manage-python-environments-pipenv.md
```diff
@@ -0,0 +1,162 @@
...
+
+        mkdir python-example && cd python-example
+
+2.  Create a virtual environment in the directory:
+
+        pipenv --python 3.6
+
+    {{< output >}}
+Creating a virtualenv for this project…
+Using /home/username/miniconda3/bin/python3.6m (3.6.4) to create virtualenv…
```
GuessWhoSamFoo: We could hold this until end of April so the install on Py3 doesn't have to rely on Miniconda

GuessWhoSamFoo:
Added missing front matter from: https://gohugo.io/getting-started/usage/#draft-future-and-expired-content

PR #1653, opened 2018-03-28
Deprecate file excerpt

Jskobos:
Since file and file excerpt are now identical, all excerpts have been converted to files. 

In addition, omitting a filename and syntax style from a file shortcode will now render the file code without any label at the top.


GuessWhoSamFoo:
Writer's formatting guide needs to be updated before this merges

PR #1655, opened 2018-03-28
Add .idea pattern to gitignore

gilwi:
Add gitignore pattern for .idea folders
Remove ci/.idea folder introduced in #1528


.gitignore
```diff
@@ -32,3 +32,4 @@ Thumbs.dbi
 *.elc
 .\#*
 *.sublime-workspace
+**.idea
```
GuessWhoSamFoo: `.idea/` should work fine or maybe `**/.idea`. The current glob pattern might be confusing.

PR #1662, opened 2018-04-01
IPv6 overview: remove incorrect/misleading section

kenyon:


cwlinode:
Thank you. We've updated the page in a different pull request.

PR #1671, opened 2018-04-04
Update NixOS guide for 18.03

nixy:
NixOS just had their 18.03 release and I went through this guide confirming that everything still works.

I didn't run into any issues, so the content is still fine I would just like to bump the modified date and version mentioned inside so people know this is maintained.


GuessWhoSamFoo:
@nixy Thanks for updating! 👍

PR #1680, opened 2018-04-10
[NEW] String Manipulation in Python 3

Jskobos:


GuessWhoSamFoo:
Looks like https://docs.python.org/3.6/library/string.html#format-specification-mini-language is left out intentionally. The document is not to be an exhaustive reference, but there should be discussion on what it is trying to accomplish (seems to be an advertisement for f strings)

PR #1683, opened 2018-04-10
Markdown parser for enforcing style and formatting

GuessWhoSamFoo:
This is the pre-alpha script to redo the awkward pytest parsing.

- [x] Move the remaining functionality from pytest over to this script
- [x] Add tests with sample markdown in `data/`
- [x] Decide whether redirect sweeps via scrapy should be handled elsewhere
- [x] Move redirect sweeps to an entirely different test
- [ ] Fix globbing with case sensitive extensions
- [ ] Define exactly what should return an exit code of 1

Would be super nice to have:

- Regex to capture image paths and check for unused assets
- Stack implementation to check for spacing on numbered lists
- Enforce `-` rather than `_` on URLs

### DIFF
--- a/ci/blueberry.py
+++ b/ci/blueberry.py
@@ -264,11 +264,15 @@ class Reporter(object):
                              self.get_yaml_error_count()
         return self.total_errors
 
+    def rule_count(self):
+        return len(list(itertools.chain(*_validate.values())))
+
     def report(self):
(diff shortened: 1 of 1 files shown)


GuessWhoSamFoo:
Remove `.pytest_cache/` and add to gitignore

PR #1688, opened 2018-04-11
Include the copyKernels grub parameter for NixOS

slimey:
The instructions failed for me with 18.03 until implementing this change

### DIFF
--- a/docs/tools-reference/custom-kernels-distros/install-nixos-on-linode.md
+++ b/docs/tools-reference/custom-kernels-distros/install-nixos-on-linode.md
@@ -137,10 +137,13 @@ The LISH console requires certain kernel and GRUB options to be configured in th
 
 ### Configure GRUB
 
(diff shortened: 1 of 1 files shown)


GuessWhoSamFoo:
Thanks for sharing this bit of information! We definitely need to find the root cause of copyKernels and failure modes before adding this to the documentation.

slimey:
OK! I don't understand the root cause, so just sharing that this workaround did it for me.

Without this, then the freshly installed partition always hung at the grub stage for me.

nixy:
So I'm planning to touch up this guide for the recent 18.09 release and using the 18.09 ISO I was able to boot without issue using the original guide (without copyKernels)

I'm not sure what problem you had, but I'm if `boot.loader.grub.copyKernels` was the solution then you may have deviated from the original guide somewhere. It's also possible there was a regression in 18.03 when you tried this that I'm not getting on the new 18.09.

I haven't been able to reproduce the error, so I think this change may be unnecessary. @slimey If you still have the .nix files you used during installation and are willing to share them I'd love to take a look to see if I could reproduce the issue you had.

slimey:
Thanks @nixy - I've just tried with 18.09 and can confirm that the instructions work for me now as-is. So this must have been an 18.03-specific issue. Thanks for checking!

PR #1689, opened 2018-04-12
Minor metadata fixes for existing guides

Jskobos:


GuessWhoSamFoo:
Add some of these cases into #1683

PR #1728, opened 2018-04-30
Update Hugo version to .40.2 in Travis

Jskobos:
Updates Hugo version used in Travis builds to 0.40.2, which provides better support for the .GetPage.Content method we use in our content shortcode.


GuessWhoSamFoo:
@Jskobos Make sure to update the Salt confs as well

Jskobos:
@GuessWhoSamFoo thanks! Turns out I can't but Brad took care of it for us yesterday.

PR #1734, opened 2018-05-01
[NEW] Use systemd to Start a Linux Service at Boot

Jskobos:


Jskobos:
Per request, we need to add a note clarifying that this guide will work on CentOS 7 (and by extension RHEL). Ubuntu/Debian also fine.

PR #1739, opened 2018-05-05
Authentication header => Authorization header in API Getting Started guide

lf-:
Fix an issue where the Authorization header was incorrectly called Authentication.


Jskobos:
Thanks for the catch!

PR #1743, opened 2018-05-07
[NEW] FreeNAS with Block Storage guide

EdwardAngert:


nmelehan:
Added link to Community site in Caution block at top and included some technical edits

PR #1744, opened 2018-05-07
[NEW] Block Storage Volumes with the Linode API

Jskobos:


docs/platform/api/create-block-storage-volumes-with-the-linode-api.md
```diff
@@ -0,0 +1,197 @@
...
+  - '[API Documentation](https://developers.linode.com/v4/introduction)'
+---
+
+<!--- ![Block Storage Volues with the Linode API](/docs/assets/api-block-storage/api-block-storage-smp.png) --->
+
+The Linode API allows you to create, delete, attach/detach, clone, and resize Block Storage Volumes.
+
+## Before You Begin
+
+You will need a Personal Access Token to access the `/volumes` endpoint. Access tokens can be created through the beta [Linode Manager](https://cloud.linode.com). See our [Getting Started with the Linode API](/docs/platform/api/getting-started-with-the-linode-api/) for more information.
```
Dorthu: If you linked [Linode Manager](#) directly to https://cloud.linode.com/profile/tokens, it might be more helpful

docs/platform/api/create-block-storage-volumes-with-the-linode-api.md
```diff
@@ -0,0 +1,197 @@
...
+
+When you have a token, store it as a temporary shell variable to simplify repeated requests:
+
+    token=<Access Token>
+
+## Create a Volume
+
+You can create a new Block Storage Volume by making a POST request to the `/volumes` endpoint.
+
+1.  Create the Volume. Adjust the size, region, and label in the request below to the desired values. You can view a list of available regions by sending a GET request to `https://api.linode.com/v4/regions`.
```
Dorthu: Can you ajax on this page?  If so, it would be fairly easy to get a dynamic list of available regions embedded in this doc.

docs/platform/api/create-block-storage-volumes-with-the-linode-api.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Detach
+
+To detach a Volume from your Linode, use the `/detach` endpoint:
+
+    curl -H "Authorization: Bearer $token" \
+    -X POST \
+    https://api.linode.com/v4/volumes/$volume_id/detach
+
+### Create and Attach
```
Dorthu: Might consider this as the first example, since it's way more useful.. maybe the "Attach" example could show you how to move a volume from one Linode to another?

docs/platform/api/create-block-storage-volumes-with-the-linode-api.md
Jskobos: Technically we can but it's usually more trouble than it's worth. After taking in your change below, the user is going to end up copying the region from the Linode they're attaching to so listing the available regions isn't necessary.

PR #1766, opened 2018-05-14
Comment out the example UDF tags

EdwardAngert:
copy of @abrining bits/854


docs/platform/stackscripts.md
```diff
@@ -266,7 +266,7 @@ The UDF tags are explained in the table below:
 |manyof   | A comma separated list of values| Optional
 
 
-Below is an example implementation of the UDF variables. You will note that the UDF tags are commented out; this is done to prevent execution errors, as the stackscript system parses the tags without removing them.
+Below is an example implementation of the UDF variables. Note that the UDF tags are commented out to prevent execution errors, as the stackscript system parses the tags without removing them:
```
Jskobos: Do we know what "parses tags without removing them" means? Maybe it's obvious to anyone working with StackScripts but it seems like maybe it should be clarified.

abrining: I feel that this is about as clear and concise as it gets. Maybe something along the lines of "reads the tags, but does not remove them from the script."

Jonchun: This is pretty standard terminology that IMO should be clear to anyone attempting to create a StackScript/work with bash scripts. What part of the statement needs to be clarified here?

PR #1769, opened 2018-05-15
Note: Inbound ICMPv6 must be allowed through firewall for pings to return

cwlinode:


Jskobos:
Closing in favor of #1772

PR #1774, opened 2018-05-16
fix header image

rmcintosh:
screen shot 2018-05-16 at 4 42 13 pm


Jskobos:
I don't know why Vale is hanging, but #1775 included this fix as well so I'm going to close this one. Thanks for the heads up @rmcintosh, I think I found all the remaining instances of this problem.

Issue #1792, opened 2018-05-20
Proposal - Reorganization of assets with page bundles


GuessWhoSamFoo:
Example in #1791 

`docs/assets/` is not manageable with unused assets, multiple conventions for naming, subdirectories, and error prone when referencing the path in markdown.

Grouping documents with [Page Bundles](https://gohugo.io/content-management/page-bundles/) provide a much cleaner way to group relevant images to a doc.

Jskobos:
* Guides converted to use Hugo bundles through the use of a Python script, with error checking and manual fixes afterwards.

* Images were copied from the old `/assets/` directory rather than moved, because of the issue of cross-referencing (multiple guides referring to the same image).

* Currently, grepping for /docs/assets/ references in the docs shows that the only remaining references are to shell scripts (accessed with wget) and config files. Once we're confident we have found all of the edge cases, we can delete all .png .jpg and .gif files from /assets/.

* Overall, both end user experience and internal procedures were basically unchanged. Some of our tests had to be modified, and one front matter variable was removed (the `image` field).

PR #1811, opened 2018-05-30
Bundle conversion - DO NOT MERGE

Jskobos:


Jskobos:
Sending up the Bat Signal on this one @GuessWhoSamFoo. What's weird here?

Guaris:
merging

Guaris:
j/k

Jskobos:
From here (assuming nothing is broken that I can't see) we still have to manually check ~245 cases where /docs/assets/ is still referred to from within guides. Most of these are shell scripts, config files, and occurrences within wget commands. Also a few images that broke the script and were skipped.

GuessWhoSamFoo:
@Jskobos 
- Duplicate assets in `/docs/assets/` needs to be addressed. Unused assets should be removed (otherwise there is little gained here imo). Even better if style guides addressed naming conventions in assets and write a few checks to enforce.

- IMAGE_LINK_REGEX (https://github.com/Jskobos/docs/blob/master/ci/blueberry.py#L37) needs to handle *.gif in matching

- Consider removing the `image` front matter var. It's linked to https://linode.com/docs/assets/*

- Some shortguides were duplicated (e.g. development/python/index.md  development/ror/index.md etc). Definitely should not have `_index.md` and `index.md` in the same directory

- Would be nice to have the decision making process documented in #1792 - Linode docs is on the Hugo showcase and any sufficiently large site will run into a similar issue.

PR #1822, opened 2018-06-01
[NEW] Getting Started with Go Packages

Jskobos:


jdkato:
Assuming you aren't already aware, the Go-related false positives in this build are due to #1666 being a little too greedy.

You'll want to change `(?s) *({{< highlight .* >}}.*?{{< ?/ ?highlight >}})` to something along the lines of `(?s) *({{< highlight [^>]* ?>}}.*?{{< ?/ ?highlight >}})`.

Here are two regex101 sessions that illustrate the issue: [before](https://regex101.com/r/E8CSLB/2) and [after](https://regex101.com/r/odWWwg/2).

Jskobos:
@jdkato thanks! Should have caught that when we were putting `{{< highlight >}}` in there in the first place. #1825 should fix it, will go in soon.

PR #1832, opened 2018-06-07
Move images from /assets to new homes

Jskobos:


nmelehan:
By moving the contribute banner image to the hercules repo, we decouple this PR and https://github.com/linode/hercules/pull/83

We're currently moving all images to be grouped with their guides, but in this case the majority of the markup for this 'guide' is in the theme. There's no Hugo method for bundling layouts and their images in the theme directory structure. But, this shouldn't be an ongoing problem, as there's only one other image referenced by the theme (the site logo in the main navigation).

PR #1845, opened 2018-06-13
Direnv2

leslitagordita:


Guaris:
Guides done, even with this nasty commit history.

PR #1853, opened 2018-06-16
fix this wording

brewn:
"Entails" isn't really the right word here since installing a custom linux distro isn't like..a necessary consequence of what the guide is trying to accomplish, it's the actual purpose of the guide. 

Also it's not *your* KVM Linode if you don't have a KVM Linode, which is the whole point of the note. 

Also, if Xen is really 100% gone it might be worth getting rid of the "If you're currently running Xen.." bit.


Guaris:
Thanks!

PR #1855, opened 2018-06-18
Manual 404 link test

leslitagordita:


GuessWhoSamFoo:
@Jskobos Why not just add `http://localhost:1313/docs/contribute/thankyou` to start_urls in scrapy?

Jskobos:
@GuessWhoSamFoo That finds 404's within the thankyou page (so it's a good idea to include it) but doesn't seem to check the url itself. We tested with a dummy URL and it doesn't detect a 404.

PR #1858, opened 2018-06-19
Add thank-you page to scrapy link targets

Jskobos:


Guaris:
I get it, but isn't that weird?

Jskobos:
A bit. There are two internal links inside the thank you page, currently not being checked by the scraper. We don't really change that page much so not very important, but coverage is coverage

PR #1864, opened 2018-06-21
Add notation for MySQL 5.7

abrining:
Had someone get tripped up by the temporary password. Not sure why 5.7 was installed on their Linode as 5.6 was installed on mine, but I figured this would be a good addition in the event that someone else does the same.


Guaris:
Awesome thanks! I updated it, and it now exists as PR #1873

PR #1872, opened 2018-06-26
fix: update to g6 plans and current distros

mchalet:
The legacy CLI [guide](https://www.linode.com/docs/platform/linode-cli/) has outdated plan and image data. This PR updates the doc to reflect current availability. 

~~To make this doc more resilient to future changes, I suggest removing the sample output for `linode configure`, `linode plans`, `linode locations`, and `linode distros`. It's easy enough to run these commands in your own terminal, and having mutable data in a doc is difficult to maintain.~~

That said, I've gone ahead and kept the current format the same, but it would be easy enough to refactor.

### DIFF
--- a/docs/platform/linode-cli/index.md
+++ b/docs/platform/linode-cli/index.md
@@ -122,25 +122,25 @@ To start with, most users will want to run the configuration utility:
 
          Default distribution when deploying a new Linode or rebuilding an existing one. (Optional)
          Valid options are:
-           1 - Arch 2017.01.01
-           2 - CentOS 5.6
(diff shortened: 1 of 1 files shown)


GuessWhoSamFoo:
Use Hugo's data template feature to make a request during build time against the `https://api.linode.com/v4/linode/types` API endpoint (likewise for distro, location, etc). Creates a single source of truth for all documentation

cc @Jskobos

mchalet:
@GuessWhoSamFoo Thanks for the tip - I'll take a look at using data templating.

Issue #1801, opened 2018-05-23
Securing your server - SSH Daemon Options (Point 3)


jamesmacwhite:
https://www.linode.com/docs/security/securing-your-server/

I'm intrigued as to what the benefit of restricting SSH to IPv4 or IPv6 does in terms of security? To me this seems a little misleading and is more security by obscurity. 

Right now you could argue that having SSH on IPv6 only would significantly cut down on the port scan attempts, given the majority will still come from IPv4, but it doesn't really provide any security and is only due to the lack of IPv6 adoption, which will _eventually_ supersede IPv4.

Perhaps the better approach is to mention using a firewall to restrict the SSH port to a specific list of hosts or IP ranges, rather than have it open to any host.

nixy:
It's 100% security through obscurity, but that doesn't mean it doesn't have value.

If you know for a fact that you'll only access your machine over IPv4 or IPv6 then doing this lets you cut out a huge swath of addresses that could be trying to login. That said, as you point out, this is pretty much a weaker version of a whitelist for trusted hosts.

A whitelist of trusted hosts is far more secure, but is also incredibly cumbersome. IMO this security guide isn't for experts, but is for people who are just starting out with servers and need to know how to protect themselves from brute force attacks. For that purpose I think this is fine.

Mentioning that a whitelist can be used to further restrict it is probably worthwhile, but it doesn't seem like a good idea to actually walk through that process in this guide. Though Linode may not want to mention something without expanding on it, as that ends up with questions to their support team.

jamesmacwhite:
@nixy. I understand what you're saying and can appreciate the guide is aimed at more novice users, but my concern is it a false sense of security the statement gives off. A port scan is absolute, doesn't really matter if its v4 or v6, same endpoint, just different IP version to get there. If a service is listening and visible on the public interface its a target. I mean, if this is a more novice guide, would some novices enough know what IPv4 and IPv6 is? Genuine question, not trying to be facetious.

My view on this remains that it should be tweaked or at least de-emphasise any real security benefit from it. Its in the same category as changing the SSH port to something like `54253`. That would cut down port scans, but it doesn't mean you've improved "security". In fact you've probably made it harder to connect on networks that restrict TCP access. Sometimes SSH 22 might be blocked, but you've got more of a chance of a standard TCP port, rather than some random higher unregistered one!

You mention more advanced SSH topics, I think if they exist, or would need to be written, references to such advanced topics from the novice guide isn't a bad idea.

[...]

PR #1876, opened 2018-06-27
Add link to OpenSSH guide

leslitagordita:


nixy:
This is definitely better, but the original issue being discussed in #1801 was using a firewall to whitelist access to port 22 by IP address. That isn't covered by the advanced guide linked.

I don't care that much (it was @jamesmacwhite that brough it up) and it is rather complicated as you would need to cover multiple firewalls types to give an in depth tutorial. Maybe a short section (paragraph or two) could be added to the advanced SSH security guide mentioning that you can use your distributions firewall to restrict access to port 22 specifically and then link to the three firewall guides (iptables, ufw, firewalld)

PR #1878, opened 2018-06-27
Fix speed up wp publish date

nmelehan:
The guide isn't appearing in the websites/cms section or in search and we think this because the publish date was set in the future when the guide was originally pushed to master


nmelehan:
This accidentally includes a commit from a draft of a new guide that's not ready to publish yet, so I'm closing this and recreating it.

PR #1887, opened 2018-07-03
Guide on some perks of developing in Linux

cwlinode:


nmelehan:
Did some significant re-writing; replaced the second section entirely, as I thought we could better sell the different methods for how someone unfamiliar with Linux could get on board.

PR #1888, opened 2018-07-03
[Update] Accounts and Passwords

leslitagordita:


nmelehan:
Replaced iframe embed of YouTube video with the Hugo shortcode for YouTube:
https://gohugo.io/content-management/shortcodes/#youtube

YouTube embeds were previously breaking the layout when the browser window was made narrow, and adopting the Hugo shortcode fixes that.

Going to make a separate PR to fix other iframe YouTube embeds in other guides.

The user permissions page in the Manager is currently missing a dollar sign next to the block storage volume grant. Will make a separate PR for that after the holiday. We will ultimately need to take another screenshot of that page for this guide and update the markdown when that website change is merged.

Issue #1915, opened 2018-07-18
docs404.py failing


nmelehan:
These issues are fixed now, but we have a summary to preserve what happened:
---

The docs404.sh script saw a number of recent interrelated issues that took a few pull requests to resolve. Here's a timeline:

### Prologue
On June 15, we observed that the Thank You page was failing the docs404 test. This was addressed in these PRs:

https://github.com/linode/docs/pull/1849
https://github.com/linode/docs/pull/1852
https://github.com/linode/docs/pull/1855
https://github.com/linode/docs/pull/1858

### New issues on July 12

Starting July 12, @Guaris observed CI tests started to result in false-positives from Travis; in other words, links that were 404s were not reporting as such. Here is a PR from that date whose docs404 test passed, but which reported the existence of 404s in the test logs:

https://github.com/linode/docs/pull/1899
https://travis-ci.org/linode/docs/jobs/403099290
```
404 response in HTML - see logs
No 404s were found from manual test
The command "./ci/scripts/$TEST_SUITE.sh" exited with 0.
Done. Your build exited with 0.
```

Later on July 12, @leslitagordita pushed a PR that fixed a broken image link in the Rocketchat guide: https://github.com/linode/docs/pull/1900

This PR was never merged, but the Travis logs show that no 404 links were found from Scrapy:
```
Scraper did not find any 404 links
No 404s were found from manual test
```

### Continuing issues on July 13

@Jskobos discovered the reason for the false-positives: in [PR 1855](https://github.com/linode/docs/pull/1855/files), we added another test and a second line to docs404.sh:

```
python ci/docs404.py
+python ci/test_urls.py
```

The bash script was judging its success by the result of the last line, which was `python ci/test_urls.py`. The fix was to combine the two lines into one: 

```
python ci/docs404.py && python ci/test_urls.py
```

This was packaged into [PR 1902](https://github.com/linode/docs/pull/1902), along with the update to the Rocketchat guide/image link that @leslitagordita also made in [PR 1900](https://github.com/linode/docs/pull/1900).

### Continuing issues on July 18

[PR 1902](https://github.com/linode/docs/pull/1902) was not merged until July 17. On July 18, @Guaris noticed that PRs were failing the docs404 test again. We found that this was because a dependency of Scrapy called Twisted was failing. The following errors were appearing in the docs404 logs in all PRs from July 13 onward:

```
Traceback (most recent call last):
  File "ci/docs404.py", line 4, in <module>
    import scrapy
[...]

PR #1917, opened 2018-07-20
Parse response code for start urls; handle shell script failure modes

SamanthaFullerton:
Closes #1912 #1911 #1907 
Addresses #1915 #1902 #1855 

#1855 was never needed - Scrapy can return the response code of starting urls.

https://doc.scrapy.org/en/latest/topics/spiders.html?highlight=parse_start_url#scrapy.spiders.CrawlSpider.parse_start_url

URL with https such as `start_url = ['https://localhost:1313/docs/invalidurl` will still fail silently and should be addressed in a separate PR.

#1902 chains the two commands together but still does not fix the breaking pattern.

e.g This would exit 0
```
python ci/docs404.py && python ci/test_urls.py
python -c "import this"
```
Failures should terminate immediately and piping to force something to succeed is discouraged.


leslitagordita:
@SamanthaFullerton Thanks for this PR!  Cleared a lot up for me.

PR #1929, opened 2018-07-26
[NEW] How to Create a Limited Access Account

EdwardAngert:


nmelehan:
Added and edited some copy to explain things to a layperson reading the guide. A lot of the time when support is asked these questions, it's from a business person with no Linux knowledge.

PR #1948, opened 2018-08-06
Use recommended installer size

jonrichards:
The official docs recommend a 300mb installer size.
(https://www.debian.org/CD/netinst/)


Guaris:
Thanks!

PR #1952, opened 2018-08-06
Theme rebuild for subnav styling fixes

nmelehan:


nmelehan:
Tested on safari, chrome, firefox at all window widths, looks the same on each 👍

PR #1955, opened 2018-08-07
[NEW] Install Redmine on Ubuntu 18.04.

Darkstar90:


andystevensname:
**Missing /etc/nginx/passenger.conf**
Step 6 in the section 'Configure NGINX,' restarting nginx, throws an error. I used `sudo nginx -t -c /etc/nginx/nginx.conf` to check the configuration, and the output referred to the non-existant `/etc/nginx/passenger.conf` file included in `/etc/nginx/nginx.conf` in step 1 of the same section:
```
nginx: [emerg] open() "/etc/nginx/passenger.conf" failed (2: No such file or directory) in /etc/nginx/nginx.conf:57
nginx: configuration file /etc/nginx/nginx.conf test failed
```
In a cursory search I found [this GitHub issue](https://github.com/phusion/passenger/issues/2076), but I'm not seeing how the user actually was able to fix the problem.

Darkstar90:
I'll take a look at this today. The issue with the Seafile guide has also
been addressed also.

kwunyeung:
I have the same issue after upgrading from 16.04LTS. I have reinstalled nginx, passenger and the libnginx-mod-http-passenger and I have checked all conf files but still not working.

PR #1967, opened 2018-08-13
[NEW] How to download a copy of a disk backup

EdwardAngert:
A collection of mostly existing guides/sections rolled into one to make downloading a copy of a backup easy.

Shortguides created:

- `mysqldump-database-backup-short`
- `mysqldump-database-restore-short`
- `copy-disk-over-ssh-short`
- `restore-backup-image-short`


nmelehan:
Undid the change to the block storage guide as they weren't needed for this PR (talked to @EdwardAngert about this first), and rolled back the vale dictionary change that came as a result of the BS guide changes.

Did a little bit of rearranging and added some new copy in the primary download backups guide. Also split out a section of the FileZilla guide into a shrotguide so it could be included in the download backups guide:

`/docs/tools-reference/file-transfer/filezilla-shortguide/`

PR #1996, opened 2018-08-23
Refactor shortguides

nmelehan:
This moves all shortguides that didn't use [leaf bundles](https://gohugo.io/content-management/page-bundles/) (e.g. `my-shortguide.md`) into leaf bundles (e.g. `my-shortguide/index.md`).

Also changes all shortguide yaml used (i.e. `shortguide: true`) to use [`headless`](https://gohugo.io/content-management/page-bundles/#headless-bundle) instead (i.e. `headless: true`). `headless` works a bit different from `shortguide`: it not only suppresses the guide from the index and the search index, it also prevents the guide from rendering on its own at all (i.e. unless embedded in another guide), so it can't be directly linked to.

Two scripts were used to do this:

1. [Convert shortguides to leaf bundles](https://gist.github.com/nmelehan/2eae45755616c1cfd0b1db68bc0c33a8)
2. [Rename `shortguide` to `headless`](https://gist.github.com/nmelehan/15bc287549f5756cb599b65e67297a0a)

This PR also updates the `shortguide` archetype and the Linode Writer's Guide to show how to use leaf bundles and the `headless` yaml for shortguides.

A large number of files were changed in this PR. The changes were verified as successful by:

1. Spot-checking that `headless` guides do not render on their own/are not directly linkable and do not appear in indexes or the search index.
2. Spot-checking that `headless` guides embed normally.
3. Verifying that the 'Content shortcode is empty' message does not appear when building Hugo locally, which is an error message given by the [`content` shortcode](https://github.com/linode/hercules/blob/master/layouts/shortcodes/content.html).


GuessWhoSamFoo:
Why bother with bundles at all if the path is abstracted away?

Hiding from RSS feed should be part of being headless

Limited user note has too much complexity for what can be a shortcode

@Guaris

nmelehan:
`headless` won't work on non-bundled guides, and bundles allow us to bundle images in shortguides as we would for any other guide.

We're renaming `shortguide` to `hiddenguide` and repurposing it for some other guides that we do want to render and be directly linkable.

I thought that `headless` by itself should hide a guide from RSS, but I'll check on that.

I prefer having another shortguide to having another shortcode (for limited user note), but maybe other members of the team disagree

nmelehan:
Hi @GuessWhoSamFoo 

Here's what I've found about why `headless` should suppress a guide from the RSS feed without needing `show_on_rss_feed` also set to false:

- [`headless`](https://gohugo.io/content-management/page-bundles/#headless-bundle) prevents a guide from ending up in .Site.RegularPages

- We use `.Data.Pages` to compile the RSS feed ([here](https://github.com/linode/docs/blob/master/themes/docsmith/layouts/rss.xml#L16))

- `.Data.Pages` is [an alias of `.Pages`](https://gohugo.io/variables/page/#pages)

- In the home context, [`.Pages` is the same as `.Site.RegularPages`](https://gohugo.io/variables/page/#pages)

- We only render RSS [in the home context](https://github.com/linode/docs/blob/master/config.toml#L31)


Can I ask if you've seen any shortguides in the RSS feed since this PR? I haven't seen any myself

PR #2009, opened 2018-08-31
[Fix] Fixed token retrieval command

junkpiano:
It must be `worker`. Otherwise we get a different result from screenshot.


Guaris:
Hey Thanks!

PR #2038, opened 2018-09-24
Rsyracuse non reboot static config

Guaris:


nmelehan:
Hey Ryan, can I ask when this comes up most for support? Is it when someone has a working static config but needs to add a second IP, and they don't want to reboot?

If so, I'd recommend removing the `route add` line, as the Linode should only have one default route, which would already exist from the first IP on the Linode. Also, you presumably wouldn't need the first `ip link set eth0` at that point either

rsyracuse:
This was specifically designed to address IP swaps, so while the `route add` line may not be necessary, the `ip link set eth0` line is (and at minimum wouldn't hurt anything). I think we can safely ignore the former though. Will get that updated

nmelehan:
Hey @rsyracuse, in the case of IP swaps, i believe you would also need to update the route (unless the new IP happened to be in the same octet?), is that right? In which case, there should also be a command for removing the prior to adding the new route, I believe (but I haven't tested this).

Perhaps we should also add a sentence which states that the instructions are meant for run-time/non-rebooting IP swaps, too? Since the commands seem fairly specific to that scenario. And, maybe we could add an extra similar bit of text and set of commands for the case of adding secondary IPs at run-time, if that question comes up often enough.

rsyracuse:
@nmelehan I tried this a couple times, and I _think_ the best way to do this would be as follows

sudo ip addr flush dev eth0
ip link set eth0 up
ip addr add 198.51.100.5/24 broadcast 198.51.100.255 dev eth0
ip route add default via 198.51.100.1  

I wasn't seeing any issues with the ARP cache, but I can see what appears to be the route issue you were describing, and a few issues that come from not flushing the interface first. 

This is testing with consistent results on my end, do you know of any other potential issues that might come up from using this?

PR #2054, opened 2018-10-04
changed LetsEncrypt link to point to Linode's guide

bennettnw2:
Changed the LetsEncrypt link from pointing to the LetsEncrypt homepage, to pointing to Linode's guide on Let's Encrypt instead


Guaris:
Boom! Nice, Thanks!

PR #2058, opened 2018-10-04
[Copy] Salt Command Line Reference

StarfallProjects:
One technical query:
In the salt-api section, you include a --user option. But this is not mentioned in the Salt documentation.
https://docs.saltstack.com/en/latest/ref/cli/salt-api.html
Should it definitely be there?


andystevensname:
There shouldn't be `--user`, this was a mistake on my part.

cwlinode:
Closed in favor of https://github.com/linode/docs/pull/2061

PR #2065, opened 2018-10-08
linux-hardened kernel support was dropped.

ellisbenjamin:
Alpine dropped support for grsec's hardened linux kernel. It is no longer part of the distribution.

see: https://git.alpinelinux.org/cgit/aports/commit/?id=94a65a421705eb0152c2a6cdeb0bffd269c58e97

from bellis@linode.com


Guaris:
Thanks!

PR #2068, opened 2018-10-09
sshd service restart - error

foranastes:
sshd should be used instead of ssh in centos (I don't know about other distros


Guaris:
Thanks a lot!

PR #2071, opened 2018-10-10
Email with postfix dovecot and mariadb on centos 7

wireframeskull:
This update largely incorporates the recently updated general guide [Email with Postfix, Dovecot, and MySQL](https://www.linode.com/docs/email/postfix/email-with-postfix-dovecot-and-mysql/), adapted to work with CentOS 7.

After testing, the comment regarding the stored location of received emails appear to be caused by a misconfiguration in the reporter's dovecot settings.

### DIFF
--- a/ci/vale/dictionary.txt
+++ b/ci/vale/dictionary.txt
@@ -888,6 +888,8 @@ ossec
 ostemplate
 osx
 otomen
+outputter
+outputters
 overcommit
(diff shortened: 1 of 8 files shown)


leslitagordita:
This PR was resubmitted as XR:https://github.com/linode/docs/pull/2076. Closing this.

PR #2078, opened 2018-10-16
[NEW] A Beginner's Guide to Salt

cwlinode:


nmelehan:
I did a pretty extensive rewrite and added a bunch of examples, which I hope makes it easier to 'see' how Salt is used. I'd ask that the final reviewer just read it once-through to pick up on any spelling errors I made.

PR #2101, opened 2018-10-30
Rebuild comm

Guaris:


nmelehan:
The shortcodes appear as expected, but it seems strange to me that we aren't placing them in the existing Join Our Community section that's at the bottom of the standard guide layout:

https://github.com/linode/hercules/blob/master/layouts/_default/single.html#L27
https://github.com/linode/hercules/blob/master/layouts/partials/community_link.html

---

If we wanted to put the links in that section, we would have to think about a different way to do it than by using a shortcode, since the section is in the theme's layout files and not in the markdown (and layout files can't use shortcodes). An alternative approach could be:

Create a new YAML option similar to the existing `external_resources` option ([documented here](https://github.com/linode/hercules/blob/380235d91925dba0b301628473edc4353dc8fd39/documentation/yaml_headers.md)) but which accepts links to the community site. Call the option `community_links` or something similar.

Then, update the `layouts/partials/community_link.html` partial to iterate through those links, similar to:
https://github.com/linode/hercules/blob/87e94c7bafae6476bab590b7d24259dd65824a54/layouts/partials/more.html

Use the same styling as in the `layouts/shortcodes/community` shortcode. Keep the CSS that was created for that shortcode. When this is all set up, delete the community shortcode file.

PR #2114, opened 2018-11-05
[Update] Getting Started with the New Manager

andystevensname:
- Updated images/gifs and copy to point to the new manager
- Gave new guide a `hiddenguide: true` and changed `show_on_frontpage: false`.
- Gave old guide a `cloud_manager_link: getting-started-new-manager`


Guaris:
Whoever copy edits can you add opensuse instructions to this?

PR #2116, opened 2018-11-06
[New] Secrets Management with Salt

andystevensname:
A new guide that discusses the following best practices for storing secrets with Salt:
- Pillars
- Environment Variables
- GPG Encryption
- SDB


nmelehan:
Fleshed out the Pillar section with a couple examples to illustrate the concepts because I thought it was the core method used for secrets and we currently don't have another dedicated secrets+Salt guide. 

Tried to keep the expansion on that section to a minimum, and refrained from expanding the other sections per Leslie's notes in the Trello card (which say that a full detailing of those methods is out of scope for this guide).

PR #2138, opened 2018-11-19
Minor edits

abrahamvegh:
- 'within' rather than 'in' because I think it’s more clear
- remove 'of course' because it talks down to the reader who may not know
- insert missing 'to'
- clarify sudo privileges
- 'on' to more appropriate 'in'


Guaris:
Thanks!

PR #2142, opened 2018-11-20
Rebuild theme - include shortguide headers in nav

nmelehan:
Hugo relies on its markdown renderer, [Blackfriday](https://github.com/russross/blackfriday), to generate the TableOfContents.

This TOC rendering step seems to happen before shortguide contents are embedded, and so any headers from an embedded shortguide did not appear in the guide navigation's TOC.

There were two options for fixing this:

- Find the code in Hugo that calls the TOC renderer and update it so that it executes after shortguides are embedded (which would be after shortcodes are rendered). I don't have a full understanding of the Hugo codebase so any such change may have unintended consequences, and any PRs for this may not be accepted.

- Re-render the TOC in Javascript after the document is ready in the browser. This option also lets us handle guides whose navs may change while the user is reading them. 
  There is only one example of that right now: the Install Git on Mac, Windows and Linux guide. The different toggle-able sections of that guide have different headers, and the articleHeadersChangedVisibility() function  is called when switching between them to show and hide headers in the nav accordingly:
https://github.com/linode/docs/pull/2142/files#diff-6dabed1cdc57820fe059dfe63808081bR285

The second (Javascript) option was chosen as the fix implementation for this PR.


themes/docsmith/layouts/partials/community_link.html
```diff
@@ -1,3 +1,3 @@
-<h3>Join our Community</h3>
+<h2 id="join-our-community">Join our Community</h2>
```
nmelehan: When this used to be a H3, it would either:

- Appear as a subsection of the 'More Information' H2 section, if the guide had 'More Information' links in the frontmatter; or:
- Appear as a subsection of the last section in the guide

The first scenario was ok, but the second scenario was not good. Promoting 'Join our Community' to an H2 to avoid this.

Also adding an ID so it can be linked to from the nav.

themes/docsmith/layouts/partials/more.html
```diff
@@ -1,5 +1,5 @@
 {{ with .Params.external_resources }}
-<h2>More Information</h2>
+<h2 id="more-information">More Information</h2>
```
nmelehan: Adding an ID so section can be linked to from the nav

themes/docsmith/static/build/js/main.js
```diff
@@ -281,23 +281,102 @@ function toggleNoteDisclosure(disclosureNote) {
...
             var $body = $(document.body);
-            var navHeight = toc.outerHeight(true) + 10;
+
+            // The y position of the sticky nav element
+            var scrollSpyOffset = 20;
 
             $body.scrollspy({
                 target: tocElemID,
-                offset: navHeight
+                offset: scrollSpyOffset
```
nmelehan: The scroll spy offset in the production site is actually incorrect: the old code calculated it from the height of the nav element. 

This lead to the scroll spy changing the highlighted section in the nav when the corresponding header was closer to the bottom of the page than the top (because the height of the nav is usually a large percentage of the height of the browser viewport).

https://getbootstrap.com/docs/3.3/javascript/#scrollspy-options

nmelehan:
Corresponding Hercules repo PR
https://github.com/linode/hercules/pull/110

PR #2148, opened 2018-11-21
[New] Install Caddy on Arch Linux

streamer45:
This guide demonstrates how to install Caddy on Arch Linux

### DIFF
--- a/docs/web-servers/caddy/install-caddy-archlinux/index.md
+++ b/docs/web-servers/caddy/install-caddy-archlinux/index.md
@@ -1,0 +1,90 @@
+---
+author:
+  name: Linode Community
+  email: docs@linode.com
+description: 'Caddy is an open source HTTP/2-enabled web server with automatic HTTPS. This guide demonstrates how to install Caddy on Arch Linux.'
+keywords: ['caddy', 'install caddy', 'archlinux', 'web server']
+license: '[CC BY-ND 4.0](https://creativecommons.org/licenses/by-nd/4.0)'
+published: 2018-10-17
+modified: 2018-10-17
+modified_by:
+  name: Linode
+title: "Install Caddy on Arch Linux"
+contributor:
+  name: Claudio Costa
+  link: https://github.com/streamer45
+external_resources:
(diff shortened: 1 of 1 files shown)


Guaris:
Hey, thanks! We are going to pull this into our normal editorial process and contact you before it gets published!

Guaris:
Hey, I had a few questions: 
What do you think about the newest licensing changes with Caddy? Does installing Caddy from the AUR like you did count as building it from source. 

Could you expand on this submission? It would be nice if the guide went into more depth about Caddy, it's features, and the benefits. We have two guides on Caddy already: https://www.linode.com/docs/web-servers/caddy/
One guide describes how to build Caddy from source, the other is similar to your guide but on Centos. When you wrote this guide did you think it would be a good addition this? 

Thanks again so much for the contribution!

streamer45:
Hi,

About the licensing changes on Caddy: I was a bit surprised myself but I understand why the developer acted that way. In the end I see it as a way to monetize from good quality open source software by providing commercial support (same idea behind Red Hat).  Also the changes don't affect the source in any way, it's only about the pre-compiled binaries.

Installing from the AUR doesn't necessarily mean you are compiling from source (many AUR packages simply download .deb equivalents and unpack), but in the specific case of my guide I am compiling from a source snapshot.

I mainly work on Arch Linux. Linode has been supporting it out of the box since I can remember but I could never find that many tutorials in the wiki that covered it.
You could argue that, given the same guide for Debian, an experience Linux developer could figure out how to adapt it to a different distribution but not always we have the time and/or patience to do that.
Also I believe Arch Linux is established enough to deserve dedicated material.
Moreover my idea was to publish a series of guides covering well-known software/technologies primarily focusing on Arch.

That said, I am more than willing to redact it and accept suggestions.

Guaris:
Thanks Claudio. I was just curious! Our team of writers & editors will begin the review process. Be on the lookout for it to be on the live site! Thanks for contributing to our open source project!

streamer45:
Great! Thank you, keep up the good work!

cwlinode:
Moving this to https://github.com/linode/docs/pull/2180.

Issue #2110, opened 2018-11-02
Vale not reliably catching spelling errors in Travis builds


nmelehan:
Vale seems to (sometimes) be passing spelling checks where there are errors present. Here's an example PR where it didn't catch `uers`:

https://github.com/linode/docs/pull/2085/files#diff-8d77b22ec6888694263d5682bb9eb011R221

After that PR was merged into develop, [develop was later merged into another feature branch](https://github.com/linode/docs/pull/2074/commits/4cac48a4ab4cf7dfbd0aa1e4f0a8417427bcfe97), and the error was caught in its subsequent PR:

https://travis-ci.org/linode/docs/jobs/449164270
Fix:
https://github.com/linode/docs/pull/2074/commits/c646c97e9ffd6e90e1d8dbed4cb4ada3b87f0fc5

(A bunch of other similar uncaught spelling errors in other guides were also caught in that same Travis build)

This happened most recently on our RC-0.2 branch:

https://travis-ci.org/linode/docs/jobs/449881835
Fix: https://github.com/linode/docs/pull/2109/commits/ffd3edbbd0312bd4548cb60bd396ddc89c0f412f
Original passing PR: https://github.com/linode/docs/pull/2101

@jdkato

jdkato:
I've done some investigating on this and it appears that:

- The problem only exists with builds `v1.0.5` and `v1.0.6` (both released on October 14). You can see that, for example, `v1.0.4` (the prior release) and `v1.1.6` (the next release) correctly identify `uers` while `v1.0.5/6` does not: [`jobs/449963172`](https://travis-ci.org/jdkato/docs/jobs/449963172).
- The reason was that `v1.0.5/6` introduced a bug (the result of attempting to fix https://github.com/errata-ai/vale/issues/84) that caused Vale to stop linting Markdown files once it encountered one starting with either a `.` or `_` (e.g., `_index.md`). [`jobs/449972873`](https://travis-ci.org/jdkato/docs/jobs/449972873) shows that, if there were no `_index.md` files, all versions correctly  identifed `uers`.

I try to ensure that Vale always follows the "stable HEAD" philosophy (meaning that the `go get` version is bug-free), but my test suite didn't have any files starting with either a `.` or `_` -- so, this slipped through my testing.

I'd like to think that this won't happen again (or atleast it'll be quite rare), but you may also want to consider pinning a version (similar to how you currently install Hugo) to gain more control over issues like this.

nmelehan:
[...]

PR #2155, opened 2018-11-28
Pin Vale to version 1.2.6

nmelehan:
For https://github.com/linode/docs/issues/2110

### DIFF
--- a/ci/scripts/vale.sh
+++ b/ci/scripts/vale.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
 
-go get github.com/ValeLint/vale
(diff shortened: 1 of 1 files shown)


ci/scripts/vale.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
 
-go get github.com/ValeLint/vale
```
nmelehan: We used to use Go to download Vale, but there currently isn't a way to use `go get` and specify a version. Instead, we now download the executable from the releases page per the repo's [README](https://github.com/errata-ai/vale#installation).

PR #2164, opened 2018-12-05
Fix button location for extra small screens

nmelehan:
#2162 created an issue where the Try This Guide button is moved to the right out of the promo code banner box:

screen shot 2018-12-05 at 4 31 19 pm

This PR fixes that by instead placing the button below the promo label and aligning it to the left (only for extra small screens):

screen shot 2018-12-05 at 4 32 04 pm

Hercules PR:
https://github.com/linode/hercules/pull/112


ci/requirements.txt
```diff
@@ -4,3 +4,4 @@ pytest>=3.3.0
 tabulate>=0.8.2
 python-frontmatter>=0.4.2
 twisted==18.4.0
+attrs>=17.4.0
```
nmelehan: Travis was showing an error like this when trying to import scrapy:
```TypeError: attrib() got an unexpected keyword argument 'converter'````

Googling the error turned up:
https://github.com/Julian/jsonschema/issues/449

Which said that older versions of the `attrs` library used an argument named `convert` instead of `converter`. That thread recommended having at least version 17.4.0 of `attrs` installed. Travis was showing that only 17.2.0 was installed (I think because one of the libraries we use listed that version as its dependency). Adding 17.4.0 as the minimum version in requirements.txt solved the issue in the next Travis builds.

PR #2201, opened 2018-12-20
Update linode-managed.md

swarlos518:
Add verbiage on what needs to be done to allow customers to choose which Linodes fall under Linode Managed.


wildmanonline:
The Managed service will apply to all Linodes on an account and will be billed at $100/mo per Linode. If this service is not needed for all Linodes, a second account can be created to separate Managed Linodes from non-Managed Linodes. Once the signup is completed for that second account, you can transfer Linodes between accounts by opening up a Support ticket requesting the transfer.

¯\_(ツ)_/¯

swarlos518:
Updated verbiage

PR #2215, opened 2018-12-31
Remove directive to disable KeepAlive for Apache

cwlinode:


nmelehan:
Looks like this was originally branched off of master, so the PR had some unrelated commits in it. I rebased this on top of develop and those unrelated commits have gone away.

PR #2222, opened 2019-01-02
[UPDATE] How to Keep Your Linode Account Safe

nmelehan:
Proposed change for DOCS-150, which is in regard to the upcoming deprecation of IP whitelisting in the classic Linode Manager. This PR inserts a note in the IP Whitelisting section which describes the upcoming deprecation, but leaves the current instructions in place.

**Do not merge** until further discussion in DOCS-150 has approved this.


Guaris:
Closed until discussion happens.

PR #2248, opened 2019-01-10
[Update] Build NGINX with PageSpeed From Source

leslitagordita:
Added two notes related to compiling from source. PageSpeed still needs to be compiled from source. There is still no support yet to install the dynamic module using a package manager on an existing NGINX installation.


dvershinin:
Thanks, but. My comment was not only to bring to attention that compiling NGINX from source is "dangerous".
But also to communicate the fact that PageSpeed module *can* be easily installed via packages.

Refer to [this post](https://www.getpagespeed.com/server-setup/nginx/install-ngx_pagespeed-dynamic-module-centos-7).

This repository is very well maintained, powered by CDN and listed in pkgs.org.
There are packages for Nginx PageSpeed module for CentOS/RHEL 6, 7 and **8**.
All packages are co-installed with SELinux policy required to run the module.

PR #2255, opened 2019-01-14
Suggest letsencrypt before self-signed

jsoref:


Guaris:
reopen this if it still is relevant

PR #2273, opened 2019-01-23
[Update] Using SSHFS To Mount Remote Directories

leslitagordita:
This PR adds fixes and updates in response to a Disqus comment.

### DIFF
--- a/docs/networking/ssh/using-sshfs-on-linux/index.md
+++ b/docs/networking/ssh/using-sshfs-on-linux/index.md
@@ -25,104 +25,96 @@ SSHFS (Secure Shell FileSystem), is a tool that allows users to securely access 
 For this guide we used two Ubuntu 16.10 systems, but `sshfs` can be installed on any Linode image.
 
 
-### Install Prerequisite Package
+### Install SSHFS
 
-Before installing SSHFS we need to update the system:
+1. Before installing SSHFS, update your system:
 
-    apt-get update && apt-get upgrade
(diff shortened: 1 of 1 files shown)


nmelehan:
The `sshfs#user@ip_address` syntax seems to be deprecated:
https://bbs.archlinux.org/viewtopic.php?id=183292

I found a new syntax for fstab that systemd can parse to create a systemd mount unit:
https://wiki.archlinux.org/index.php/SSHFS

I found that the key arguments to get the file permissions right were `allow_other` and `default_permissions` together:
https://superuser.com/questions/964821/translate-permissions-with-as-sshfs-remote-mount

Some other arguments listed in the Arch Linux SSHFS article produced what I thought of as unwanted results: 

- `idmap=user` changed the owner of files in the client mounted directory to `root` instead of whatever the user was on the remote system
- The `uid` option changed it back to whatever you set it to, but it also seemed to override the actual user that owned a file on the remote system (e.g. if a file was owned by `user2` on the remote, and the `uid` was set to `user1` in your client's fstab, then the client's mounted directory would show `user1` as the owner.

`default_permissions` and `allow_other` seemed to keep the file ownership/permissions as they were on the remote system.

Other options for FUSE:
http://man7.org/linux/man-pages/man8/mount.fuse.8.html

---

I also changed the example remote IP to 192.0.2.4, because .0-.3 is usually reserved on our infrastructure for any given /24

PR #2284, opened 2019-01-28
Rebuild theme: Add config for GitInfo.AuthorDate support

leslitagordita:
This is the theme rebuild for Hercules PR: https://github.com/linode/hercules/pull/118


nmelehan:
Testing and updates:

- I created a new staging server to test that the standard `gulp publish` flow still works and doesn't trip over the .GitInfo object. It works normally; for example, I removed the modified date from the front matter of the Getting Started guide (on a test branch). The staging server publishes normally, and the guide shows today's date in the Updated label:
http://45.33.75.101/docs/getting-started/

- I updated the tests to make `modified` not required, and then updated the Writer's Formatting Guide and the YAML documentation in the Hercules repo accordingly. I then tested that Travis still worked when the `modified` tag was absent from the front matter of a guide:
https://github.com/linode/docs/pull/2288

- I observed that when running the `gulp dev` command, the modified date from Git isn't used. As best I can tell, this is because `gulp dev` runs the hugo server from the hercules directory, with the content directory set to `../docs/docs`. If the `modified` date is still present on a guide, `gulp dev` will fall back to displaying that. If `modified` is missing, then `gulp dev` will show the original publish date in the Updated label on a guide.

    I tried to fix this, but I was unsuccessful: https://github.com/nmelehan/hercules/commit/6c846372645e2e1b8f669e2473cbe9987eea0ab8

    The idea is to set the `--source` directory to be the docs directory. This is what the `gulp publish` command does, and it works as expected (in that the .GitInfo object is accessible). In addition to setting the `--source` directory, I use the `--themesDir`, `--theme`, and `--config` options to specify the themes directory, theme name, and location for config.toml so that they refer back to the hercules directory. This part doesn't work; when hugo builds, it says it can't find various shortcodes. 

    I tried also specifying the `--layoutsDir` option, which succeeded in getting things to build, but the CSS/javascript/all styling assets were missing from the built site. I abandoned working on this, as accepting that `gulp dev` will show an inaccurate date is probably an acceptable compromise for now.

PR #2303, opened 2019-02-04
fix slash for etc

massdest:


Guaris:
Thanks ! I'll merge this in shortly!

PR #2311, opened 2019-02-07
[UPDATE] Set up a Hardened OpenVPN Server on Debian 9

andystevensname:
Adds the rule: `-A FORWARD -i tun0 -j ACCEPT` to the TUN interface rules. This was suggested to us by a reader. This is in line with the iptables rules laid out within our Tunnel Your Internet Traffic Through an OpenVPN Server guide.


andystevensname:
This was unnecessary, as the second part of this two-part guide covers this aspect of setting up a VPN.

PR #2346, opened 2019-02-26
[update]fix disk resize link

TheSchwa:
This should be pointing to the disk resize guide, not the linode resize guide.


TheSchwa:
@nmelehan fixed conflict (from classic --> new renaming) works locally. Just force pushed over my branch, hopefully didn't result in any git weirdness.

PR #2389, opened 2019-03-18
[How to Create a Private Python Package Repository] setup.py Syntax Fix

LopezNathan:
Added missing commas resulting in a syntax error when executing `setup.py`.

Article - https://www.linode.com/docs/applications/project-management/how-to-create-a-private-python-package-repository/


Guaris:
Thanks a lot!

PR #2422, opened 2019-03-31
[New] Use HashiCorp Vault to Manage Secrets

tylerjl:
Hey Linode team! Here's a new guide walking through deploying Vault. The steps here are a blend of HashiCorp's best practices and my own experiences running Vault in production, and I've walked through the guide a few times on different distributions with success so I think it's ready for review.

Note that I added a minor patch to the Certbot guide. I ran into an issue wherein `nginx` was not present on CentOS 7, so this fixes the issue going forward (I'm not sure why the certbot package isn't correctly marking `nginx` as a dependency, but this change should be safer in case there's any upstream changes in the future.)

/cc @Guaris for visibility


rsyracuse:
Thanks Tyler! We've created a new PR for this and will have it available soon. 

https://github.com/linode/docs/pull/2463

PR #2430, opened 2019-04-02
[New]Deploy WireGuard with One-Click Apps

leslitagordita:
- New guide for WireGuard One-Click App


nmelehan:
Added several steps that I thought were missing from the instructions, but I think the script has some problems still, and I haven't actually gotten a working WireGuard connection out of using the stackscript. Documented these issues and passed theme along to the one-click testing team.

PR #2443, opened 2019-04-12
[Object Storage][NEW] How to Use Object Storage

andystevensname:
**Do not merge until EAP.**

Will require updates to include static site guide links, correct Beta URL subdomains, and future changes the the CLI plugin and its download link.

- Create a guide on how to use Object Storage using Linode CLI, s3cmd, and Cyberduck.


hzoppetti:
CT-513

PR #2453, opened 2019-04-17
[update] adds dnssec section

tbaka:
this is simply to make explicit that the linode nameservers don't support dnssec and that it will break name resolution if they enable it at the registrar.


Guaris:
LGTM

PR #2464, opened 2019-04-25
Update NixOS guide for 19.03

nixy:
Updating this guide for NixOS's 19.03 release. I tested it to confirm it still works and that the guide's direction are still valid using the shiny new Linode Manager (which looks really nice and is a joy to use).

I didn't run into any issues as most of the direction on how to do things in the Linode Manager were hyperlinked to other guides that have been updated. So the content is still fine I would just like to bump the modified date and version mentioned inside so people know this is maintained.


Guaris:
Thanks LGTM!

PR #2513, opened 2019-05-31
[new] Using the context Go package'

Guaris:


docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+If you still need to install Go, you can follow our guide for Ubuntu installation [here](https://www.linode.com/docs/development/go/install-go-on-ubuntu/).
+
+{{< note >}}
+This guide is written for a non-root user. Depending on your configuration, some commands might require the help of `sudo` in order to get property executed. If you are not familiar with the `sudo` command, see the [Users and Groups](/docs/tools-reference/linux-users-and-groups/) guide.
+{{< /note >}}
+
+## About the context package
+
+The main purpose of the `context` package is to define the `Context` type and
+support *cancellation*. This happens because there are times where you want to
```
jnschaeffer: This is partly correct, but not the full story: The `context` package supports both the handling of multiple concurrent operations and passing of (typically request-scoped) contextual data in key-value pairs.

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+abandon what you are doing but it an elegant way. However, it would be very helpful
+to be able to include some extra information about your cancelling decisions and propagate
```
jnschaeffer: Not necessarily information about cancelling decisions! [This](https://blog.golang.org/context) Go blog post on contexts highlights the usefulness of `Context` objects as sets of key-value pairs for request-scoped values. Their use goes beyond providing information about why an operation was terminated (and in fact, given that cancellation occurs "above" the context that gets cancelled, it's not really clear that that would be a valid use case).

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+the information in a standard and portable way. The `context` package defines the
+`context.Context` type that can carry deadlines and cancellation signals between processes.
+
+If you take a look at the source code of the `context` package, you will realize that its
+implementation is pretty simple — even the implementation of the `Context` type is pretty
+simple, yet the `context` package is very important.
```
jnschaeffer: No need to emphasize that `context` is important, IMO - it's already been added as part of many stdlib functions. That much should be apparent to the audience.

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+{{< /note >}}
+
+## About the context package
+
+The main purpose of the `context` package is to define the `Context` type and
+support *cancellation*. This happens because there are times where you want to
+abandon what you are doing but it an elegant way. However, it would be very helpful
+to be able to include some extra information about your cancelling decisions and propagate
+the information in a standard and portable way. The `context` package defines the
+`context.Context` type that can carry deadlines and cancellation signals between processes.
```
jnschaeffer: This sentence restates the opening paragraph sentence, but more importantly is closer to what `context` actually does: carrying information about process state across goroutines.

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+
+If you take a look at the source code of the `context` package, you will realize that its
+implementation is pretty simple — even the implementation of the `Context` type is pretty
+simple, yet the `context` package is very important.
+
+The `Context` type is a Go *interface* with four methods, named `Deadline()`, `Done()`,
```
jnschaeffer: For brevity, you could also paste the definition of `context.Context` here. Go programmers are used to that. The Go blog post linked previously does that, for example.

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+`Err()`, and `Value()`. The good news is that you do not need to implement all of these
```
jnschaeffer: Not only is it good news that `Context` users do not need to implement their own contexts, but it's also not expected or even desired that they would do so. Placing this idea in the reader's mind may confuse them and make them think that's typically the case when writing Go when it isn't.

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+functions of the `Context` interface – you just need to declare and modify a `Context`
+variable using functions such as `context.WithCancel()`, `context.WithDeadline()` and
+`context.WithTimeout()`.
+
+The source code of the `context` package includes the following information:
```
jnschaeffer: Rather than provide the documentation as-is, why not help explain what this means to the reader? The reader - presumably someone with some Go experience, if not a lot - may not immediately understand the parent-child relationship between contexts and what it means for a context to be derived from another, or why you can still have resource leaks in a garbage-collected language (one fun artifact of concurrency as a language feature).

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+{{< /note >}}
+
+## A simple example
+
+The first code example will be relatively simple and illustrate the use of the
+`context.Context` type with the help of `simple.go`.
+
+### Explaining the Go code of the Example
+
+The Go code of `simple.go` is as follows:
```
jnschaeffer: `The Go code` reads awkwardly here.

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+        "context"
+        "fmt"
+        "os"
+        "strconv"
+        "time"
+)
+
+func f1(t int) {
+        c1 := context.Background()
+        c1, cancel := context.WithCancel(c1)
```
jnschaeffer: Maybe it's me being ornery, but since the explanation of `WithCancel` was included above, why not add documentation highlighting its use?

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+        defer cancel()
+
+        go func() {
```
jnschaeffer: Same as above re: documentation. A high-level explanation of what `f1` does in the code sample might be nice.

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+        }
+        fmt.Println("Delay:", delay)
+
+        f1(delay)
+        f2(delay)
+        f3(delay)
+}
+{{< /file >}}
+
+The program contains four functions including the `main()` function. Functions `f1()`,
```
jnschaeffer: This feels like a statement of what should be obvious to the reader.

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+`f2()` and `f3()` require just one parameter, which is a time delay, because everything
+else is defined inside the function.
+
+The first thing to remember is that you need to call the `context.Background()` function
```
jnschaeffer: This is good information!

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+in order to initialize an empty `Context`. The other function that can create an empty
+`Context` is `context.TODO()` and is going to be presented later in this guide.
+
+Notice that the `cancel` variable in `f1()`, which is a function, is one of the return values of
```
jnschaeffer: Good explanation, would be even better in summarized form in the code for readers who might blindly paste this in without getting this far.

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+`context.CancelFunc()`. The `context.WithCancel()` function uses an existing `Context` and creates a
+child of it with cancellation. The `context.WithCancel()` function also retuns a `Done` channel
```
jnschaeffer: Explaining the child part here is good, and would also be good to do where the documentation from the `context` source was included.

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+that can be closed, either when the `cancel()` function is called, as shown in the preceding code,
+or when the `Done` channel of the parent context is closed.
+
+{{< note >}}
+One of the return values of `Context.Done()` is a Go channel, because otherwise you would not have
+been able to use it in a `select` statement.
```
jnschaeffer: This is true of course, but why? `select` can trip some new Go programmers up because it looks like `switch` in some ways but functions differently.

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+        delay int = 5
+        w     sync.WaitGroup
+)
+
+type myData struct {
+        r   *http.Response
+        err error
+}
+
+func connect(c context.Context) error {
```
jnschaeffer: Now might be a good time to highlight that in packages that use contexts, convention is to pass them as the first argument to a function.

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+                        return err
+                }
+                defer resp.Body.Close()
+
+                realHTTPData, err := ioutil.ReadAll(resp.Body)
+                if err != nil {
+                        fmt.Println("Error select:", err)
+                        return err
+                }
+                fmt.Printf("Server Response: %s\n", realHTTPData)
```
jnschaeffer: Why not use the `log` package here? It's sort of a minor point, but helps readers get in the habit of privileging `log` over `fmt` for debugging.

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+        select {
+        case <-c.Done():
+                tr.CancelRequest(req)
+                <-data
+                fmt.Println("The request was cancelled!")
+                return c.Err()
+        case ok := <-data:
+                err := ok.err
+                resp := ok.r
+                if err != nil {
```
jnschaeffer: +1 for actually including error handling in the code; lots of Go examples skip it for brevity and encourage poor habits in readers.

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+        ctx := context.WithValue(context.Background(), myKey, "mySecretValue")
+        searchKey(ctx, myKey)
+
+        searchKey(ctx, aKey("notThere"))
+        emptyCtx := context.TODO()
+        searchKey(emptyCtx, aKey("notThere"))
+}
+{{< /file >}}
+
+This time we create a context using the `context.TODO()` function instead
```
jnschaeffer: `the context.TODO() function instead...` reads awkwardly - it's already clear it's a function, after all. Why not just `context.TODO() instead...`?

docs/development/go/go-context.md
```diff
@@ -0,0 +1,468 @@
...
+Exiting...
+{{< /output >}}
+
+Notice that `http://localhost:8001` uses a custom made HTTP server that returns
+a small amount of data. However, nothing prohibits you from trying commands such
+as `go run http.go https://www.linode.com/ 10` .
+
+## Another use of Context
+
+In this section of the guide you are going to learn how to pass values in a `Context`.
```
jnschaeffer: Good to highlight this, but with three caveats:
- The title is kind of awkward. `Another use of Context` sounds vague, when in reality this is about using contexts as key-value stores.
- The section is pretty brief compared to the others. Explaining _why_ a user might want to do this would be good.
- This section conflicts with the introductory statement that passing values into contexts is done to provide further information about why they were cancelled.

docs/development/go/go-context.md
mactsouk: Used your description of `context`.

docs/development/go/go-context.md
mactsouk: Removed that information.

docs/development/go/go-context.md
mactsouk: Added!

docs/development/go/go-context.md
mactsouk: Removed that.

docs/development/go/go-context.md
mactsouk: Removed the documentation and added an explanation about `resource leaks in a garbage-collected language`.

docs/development/go/go-context.md
mactsouk: Changed to just `code` - good point.

docs/development/go/go-context.md
mactsouk: Done.

docs/development/go/go-context.md
mactsouk: Some people might read that explanation first instead of the Go code :)

docs/development/go/go-context.md
mactsouk: Done.

docs/development/go/go-context.md
mactsouk: I think it is good as it is.

docs/development/go/go-context.md
mactsouk: Did that in the `context` introduction.

docs/development/go/go-context.md
mactsouk: `select allows a goroutine to wait on multiple communications operations`

docs/development/go/go-context.md
mactsouk: Done.

docs/development/go/go-context.md
mactsouk: That is a good point - I want the reader to be able to see the output. I will include what you said as you are right about it.

docs/development/go/go-context.md
mactsouk: Done.

docs/development/go/go-context.md
mactsouk: - Changed the title
-  The example is brief so the section is brief as well
- You are right about that :)

docs/development/go/go-context.md
jnschaeffer: Totally fine. These are all just suggestions based on my first reading, really.

docs/development/go/go-context.md
jnschaeffer: Specifically multiple send and receive operations on channels; it (somewhat confusingly IMO) doe not work for operations that may simply block.

mactsouk: I will change that - thanks!

PR #2545, opened 2019-06-13
Feature/objectstorage

andystevensname:
This contains the 
- How to Use Object Storage 
- How to Host a Static Site with Object Storage
- Some guides that were failing due to a new Vale rule (markdown > Markdown)


rsyracuse:
https://github.com/linode/docs/pull/2563

We are merging this in through a hotfix instead of this branch so we are closing it.

PR #2560, opened 2019-06-21
[Update] Automatically Configure Servers with Ansible and Playbooks & Learn How to Install Ansible and Run Playbooks

leslitagordita:


docs/applications/configuration-management/getting-started-with-ansible/index.md
```diff
@@ -0,0 +1,240 @@
...
+* Create two Linodes to manage with Ansible and establish a basic connection between the control node and your managed nodes. The managed nodes will be referred to as `node-1`, and `node-2` throughout the guide.
+
+    {{< note >}}
+The examples in this guide provide a manual method to establish a basic connection between your control node and managed nodes as a way to introduce the basics of Ansible. If you would like to learn how to use Ansible's [Linode module](https://docs.ansible.com/ansible/latest/modules/linode_v4_module.html) to automate deploying and managing Linodes, see the [How to use the Linode Ansible Module to Deploy Linodes](/docs/applications/configuration-management/deploy-linodes-using-ansible/). The guide assumes familiarity with Ansible modules, Playbooks, and dynamic inventories.
+    {{</ note >}}
+
+## Before You Begin
+
+{{< caution >}}
+This guide's example instructions will create up to three billable Linodes on your account. If you do not want to keep using the example Linodes that you create, be sure to [delete them](#delete-a-cluster) when you have finished the guide.
```
hzoppetti: I added "up to" three billable here because we say that they could 1) run the following with just one managed node and 2) they could make their home computer the control node.

docs/applications/configuration-management/getting-started-with-ansible/index.md
```diff
@@ -0,0 +1,240 @@
...
+
+If you remove the resources afterward, you will only be billed for the hour(s) that the resources were present on your account. Consult the [Billing and Payments](/docs/platform/billing-and-support/billing-and-payments/) guide for detailed information about how hourly billing works and for a table of plan pricing.
+{{< /caution >}}
+
+1. [Create three Linodes](/docs/getting-started/#create-a-linode) running Debian 9. One will be the **control node** and two you will use as your Ansible **managed nodes**. The examples in this guide can also be followed using a single managed node, if preferred.
```
hzoppetti: Here it said just to create 3 and then they would all be managed nodes, which is not quite true. Only 2 would be used for managed nodes, named node-1 and node-2, as stated above. So I clarified that.

PR #2592, opened 2019-07-15
[New] A Beginner's Guide to LXD: Setting Up a Reverse Proxy to Host Many Websites in Containers

simos:
Added guide to setup a reverse proxy in LXD in order to host many websites, each in its own container.
All websites are HTTPS (Let's Encrypt), with automated certificate updates.

Reverse proxy: nginx

Web server software: Apache2, nginx


hzoppetti:
Hi Simos - Quick question for you as I'm editing...
For these websites to work, will you need to set-up DNS to this server for these two domains? This is not included in the guide and seems to be a missing element. When setting up the web servers in the reverse proxy I don't understand how to access the sites without this.

simos:
> Hi Simos - Quick question for you as I'm editing...
> For these websites to work, will you need to set-up DNS to this server for these two domains? This is not included in the guide and seems to be a missing element. When setting up the web servers in the reverse proxy I don't understand how to access the sites without this.

The user indeed needs to set-up DNS for two domains (or hostnames). 
They are two just so that to demonstrate both Apache and Nginx Web servers. 

In the document there is this text,

_2. In this tutorial, we will be using the hostnames `apache1.example.com` and `nginx1.example.com` for the two websites that we will be creating. Create two hostnames using domains that you own and set them to the IP address of the server that you have just created in Step 1._

The domain *.example.com is reserved (by ICANN?) to be used as examples in technical documentation.

hzoppetti:
Thank you!

hzoppetti:
One more (hopefully) question for you, Simos. When I install LXD via the snap stable using Ubuntu 18.04 and then do a snap list, it lists lxd version 3.16. But, if I do an lxd --version, it says 3.0.3. 

So when I try to run the following command I get an error because it doesn't support the proxy_protocol: 

lxc config device add proxy myport443 proxy listen=tcp:0.0.0.0:443 connect=tcp:127.0.0.1:443 proxy_protocol=true 

"Error: Invalid devices: Invalid device configuration key for proxy: proxy_protocol"

Any suggestions?

hzoppetti:
Ah! Simos, please disregard my last comment - I ran a lxd.migrate to upgrade to the latest version. Although, I'm not sure why the snap didn't do this. I'm guessing I was actually running two versions of lxd? I don't know where snap puts it's version and why it doesn't overwrite the other.

simos:
> Ah! Simos, please disregard my last comment - I ran a lxd.migrate to upgrade to the latest version. Although, I'm not sure why the snap didn't do this. I'm guessing I was actually running two versions of lxd? I don't know where snap puts it's version and why it doesn't overwrite the other.

Indeed, the Ubuntu 18.04 image has preinstalled LXD 3.0.x (LTS version of LXD, supported along with Ubuntu 18.04 LTS until 2018+5 = 2023). While LXD 3.0 has support for _proxy devices for TCP<->TCP_, it does not have the `proxy_protocol` support that was added later.

Ubuntu 18.04 is the last version that has a DEB package of LXD. Subsequent versions of Ubuntu can only get the official snap package of LXD.
Considering that Linode shows by default the latest version of Ubuntu (currently 19.04) and not the LTS version, I decided to focus on the snap package and the latest stable version of LXD (currently 3.16). Because without support for `proxy_protocol`, it would be necessary to use `iptables` rules, then saving those `iptables` rules, complicating further the whole process. 

Having said that, if you have Ubuntu 18.04 LTS that already has installed LXD 3.0 from the DEB package, and then install the snap package of LXD, you end up with two installations of LXD.
The snap package installation of LXD should not (and probably  cannot) invoke automatically `lxd.migrate` because snap package installations are non-interactive, and `lxd.migrate` has to be interactive.

The LXD configuration for the DEB package is at `/var/lib/lxd`, while for the snap package is at `/var/snap/lxd/`. The executables for the DEB package are in `/usr/bin` (takes precedence), and those for the snap package in `/snap/bin`.

Here is some sample text that explains how to migrate from the DEB package of LXD to the snap package. 

> 
> If you have just created a VPS based on Ubuntu 18.04 LTS, then your system already has LXD 3.0 installed as a DEB package. For this tutorial, we are using a feature only found in newer versions of LXD. Therefore, we replace the preinstalled LXD 3.0 with the snap package of LXD. There are two ways to do so.
> 
> The first way is to unconditionally remove the DEB package of LXD 3.0. The first command below removes the main LXD DEB package. The second command removes any other LXD dependencies, including the LXD client DEB package. 
> 
>      sudo apt remove lxd
>      sudo apt autoremove
> 
> The second way is to install the snap package of LXD, and then use the provided `lxd.migrate` utility to migrate any configuration from the DEB installation of LXD into the snap package installation of LXD. By doing so, `lxd.migrate` will migrate configuration and containers (if any). Then, it will prompt you to remove the DEB packages of LXD. Select to remove the DEB packages of LXD.
> 
>      sudo lxd.migrate

hzoppetti:
Thanks so much!

docs/applications/containers/beginners-guide-to-LXD-reverse-proxy/index.md
```diff
@@ -0,0 +1,668 @@
...
+  - '[Proxy Protocol](https://www.haproxy.com/blog/haproxy/proxy-protocol/)'
+  - '[TLS Termination Proxy](https://en.wikipedia.org/wiki/TLS_termination_proxy)'
+---
+
+## Introduction
+
+[LXD](https://linuxcontainers.org/lxd/) (pronounced "Lex-Dee") is a system container manager build on top of Linux Containers (LXC) and is supported by [Canonical](https://canonical.com). The goal of LXD is to provide an experience similar to a virtual machine but through containerization rather than hardware virtualization. Compared to Docker for delivering applications, LXD offers nearly full operating-system functionality with additional features such as snapshots, live migrations, and storage management.
+
+In this guide you will:
+
```
hzoppetti: I shortened this section into a bulleted list and made it into this new format that we started using "In this guide you will:" with links to the various sections.

docs/applications/containers/beginners-guide-to-LXD-reverse-proxy/index.md
```diff
@@ -0,0 +1,668 @@
...
+### Before You Begin
+
+1.  Complete the [A Beginner's Guide to LXD: Setting Up an Apache Webserver In a Container](/docs/applications/containers/beginners-guide-to-lxd/) guide. At the end, the guide instructs you to create a container called `web` with the Apache web server for testing purposes. Remove this container by running the following commands.
+
+        lxc stop web
+        lxc delete web
+
+    {{< note >}}
+For this guide LXD version 3.3 or later is needed. Check the version with the following command:
```
hzoppetti: I added a version note about needing 3.3 with instructions about upgrading if needed as per our discussion in the comments.

docs/applications/containers/beginners-guide-to-LXD-reverse-proxy/index.md
```diff
@@ -0,0 +1,668 @@
...
+
+1.  Update the package list in the `apache1` container.
+
+        sudo apt update
+
+1.  Install the package *apache2* in the container.
+
+        sudo apt install -y apache2
+
+1. Create the file `/etc/apache2/conf-available/remoteip.conf`.
```
hzoppetti: Here, and in the Nginx section as well, I broke this out into a file shortcode.

docs/applications/containers/beginners-guide-to-LXD-reverse-proxy/index.md
```diff
@@ -0,0 +1,668 @@
...
+
+        sudo a2enmod remoteip
+
+    {{< output >}}
+Enabling module remoteip.
+To activate the new configuration, you need to run:
+systemctl restart apache2
+{{< /output >}}
+
+1. Edit the default web page for Apache to make a reference that it runs inside a LXD container.
```
hzoppetti: Here, and for the Nginx section as well, I moved this here and explicitly made this a step to complete instead of implying that it was done.

docs/applications/containers/beginners-guide-to-LXD-reverse-proxy/index.md
```diff
@@ -0,0 +1,668 @@
...
+        lxc config device add proxy myport443 proxy listen=tcp:0.0.0.0:443 connect=tcp:127.0.0.1:443 proxy_protocol=true
+
+    {{< output >}}
+Device myport80 added to proxy
+Device myport443 added to proxy
+{{< /output >}}
+
+    The `lxc config device add` command takes as arguments:
+
+    | Argument | Explanation |
```
hzoppetti: I made this a table for easier readability.

docs/applications/containers/beginners-guide-to-LXD-reverse-proxy/index.md
```diff
@@ -0,0 +1,668 @@
...
+                proxy_set_header X-Real-IP $remote_addr;
+                proxy_pass http://apache1.lxd;
+        }
+
+        real_ip_header proxy_protocol;
+        set_real_ip_from 127.0.0.1;
+}
+{{</ file >}}
+
+    You can run `sudo nano /etc/nginx/sites-available/apache1.example.com` to open up a text editor to add the configuration. Note, in this case you only need to edit the `server_name` to be the hostname of the website.
```
hzoppetti: I did not need to edit the proxy_pass field and so I removed that from this text.

PR #2596, opened 2019-07-16
[new] Learning to use lsof at its full potential

mactsouk:


andystevensname:
@mactsouk ,

I just finished a technical edit of this guide. There's a lot of great information in here, you really present a knowledgeable understanding (as you do with all of your guides). We're excited to add this information to our library!

I've done a tech edit on a number of your guides so far, and I've come up with a few areas where these guides could be improved that I thought I'd pass your way for consideration:

- When you create a guide, name the the markdown file `index.md` so that the URL structure is correct. For example, in this guide the file was named `lsof.md`, which caused the URL to be `../docs/networking/diagnostics/lsof/lsof`. That extra `/lsof` can be avoided by naming the file `index.md`.
- Your guides are chock full of examples, which is great!, but we think there could be a stronger narrative thread running through the guide that takes the reader from one step to the next. Check out my commits, you'll see that I reorganized the guide to group examples together by command option/purpose in an attempt to lead the reader logically through the examples. Don't be afraid to write a bit more to set up each example. The places where you do that (pointing out examples that are perfect/common use cases, for instance) work out really nicely.
- There are some examples that I've pulled to the top of the guide, as they seem to be foundational examples that other examples build upon. Also, I've grouped all other examples under the *Additional Examples* heading, just to give them some added context. I'd be interested to hear your thoughts on this strategy.
- Don't be afraid to nest examples in sections that seem to relate to one another. I don't think every section needs to be its own level two heading (##), and instead some could be converted to level three headings (###), or even level four.

I'd love to hear your thoughts on these suggestions. Like I said, this is all good information, and we're really glad to have it in our library. We'd love to work with you more closely to bring the best guide we can to the public.

Thank you,

- Andy

mactsouk:
@andystevensname,

I totally agree with your suggestions.

- I will use `index.md` from now on when submitting a guide.
- I will arrange the examples _better_.
- The `Additional Examples` heading is a good name that I will use from now on.
- Less level two heading is also a good idea.

Thank you!
- Mihalis.

nmelehan:

Built with commit e8a604cdabb1532740d88d2b9439ca2556ebd44f

https://deploy-preview-2596--nostalgic-ptolemy-b01ab8.netlify.com

PR #2606, opened 2019-07-18
[update] the anchor text in the link

Rajakavitha1:
The link text reads: rotating list of suggested topics.
Updated the text to: Contribute to our Tutorials

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@
 
 ## How can I write a guide?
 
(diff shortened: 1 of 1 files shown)


Guaris:
What do you think about my change?

Rajakavitha1:
> What do you think about my change?
 I love what you have. 

My two cents: From accessibility point of view . The screen reader will read 'link xyz' and the page that opens will be 'abc'.

PR #2666, opened 2019-08-18
[update] Changed to click on advanced tab rather than settings tab

bennettnw2:
Currently is says to select settings and then select advanced from there.  However, the advanced settings must have been changed to their own tab.  Which I like as it does not bury the options so deep.  Anywho, the picture will also need to be changed so that you can see the advanced tab.


Guaris:
Thanks!

PR #2699, opened 2019-08-30
An introduction to bash shell programming - part 3

mactsouk:


nmelehan:

Built with commit b0c9945592aae70d7b0bb1153dc62b0749305ae1

https://deploy-preview-2699--nostalgic-ptolemy-b01ab8.netlify.com

hzoppetti:
Hi Mihalis - You use this or similar to test for integer:
[ $arg -eq $arg 2>/dev/null ] 
several times...but I find this to be "hacky" and confusing. Is there something else you can replace this with? perhaps a regex?
This is on lines: 201, 349, and 586
Thanks!

mactsouk:
@hzoppetti Sorry, just saw your comment.
`[ $arg -eq $arg 2>/dev/null ]` is not that difficult to understand. You compare each argument with itself using `-eq`, which works for valid numbers and you ignore any error messages using `2>/dev/null`. Does it make sense now?

PR #2722, opened 2019-09-13
Clarify cloned key section

bkmdev:
1.  I had a dupe Longview key issue but didn't get the error message shown.   Re-installing by itself doesn't seem to do anything either if the client is already up to date, so it seems one may need to first *uninstall* then re-install.  

2. I also assumed the GUID shown in the URL was the same as the Longview key, which does not seem to be the case so adding a note about this may help to clarify things.

### DIFF
--- a/docs/platform/longview/longview/index.md
+++ b/docs/platform/longview/longview/index.md
@@ -294,7 +294,9 @@ If you clone a Linode which has Longview installed, you may encounter the follow
 Multiple clients appear to be posting data with this API key. Please check your clients' configuration.
 {{< /output >}}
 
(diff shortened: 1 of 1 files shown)


nmelehan:

Built with commit ab964587acbec6ca4659807f32b2e346c56878a6

https://deploy-preview-2722--nostalgic-ptolemy-b01ab8.netlify.com

bkmdev:
Thanks @nmelehan!  Markdown looks updated on the preview deployment :+1:

PS:  not sure what the CI travis build failure is about though, the test suite script seems to be returning 1:  `The command "./ci/scripts/$TEST_SUITE.sh" exited with 1.`([see here](https://travis-ci.org/linode/docs/jobs/584662279#L279))

Guaris:
@bkmdev  This should pass now

PR #2760, opened 2019-09-30
Preserve whitespace in output shortcode

nmelehan:
The `{{< output >}}` shortcode currently doesn't preserve whitespace, and also causes line-breaks when text wraps, which can end up with some poor formatting for commands that print tables, like netstat:

netstat-whitespace-pre-line

This update will preserve the whitespace:

netstat-whitespace-pre

This shortcode is being used a lot by @mactsouk in contributed guides (as it should be 👍), so this will help in reviewing those contributions. This has also come up in the past; my hack used to be to put the output into a normal tab-indented code block, and now we won't have to do that.


nmelehan:

Built with commit ef258888a4bdc0ae8c864b72c6339add6c91d645

https://deploy-preview-2760--nostalgic-ptolemy-b01ab8.netlify.com

Issue #2769, opened 2019-10-06
Old (v2.2) Access Control on Apache Config on Use NGINX as Front-end Proxy 


bagasme:
## Guides Affected
Use NGINX as Front-end Proxy and Software Load Balancer

## Issue Description
Apache site configuration on affected guide contain access control syntax from Apache HTTPD v2.2:
```ini
  ...
 <Directory /var/www/html/example.com>
   Options Indexes FollowSymLinks MultiViews
   AllowOverride None
   Order allow,deny
   allow from all
 </Directory>
  ...
```

On systems with Apache HTTPD v2.4, the access control should be (as per [upstream upgrading documentation](http://httpd.apache.org/docs/2.4/upgrading.html)):
```ini
  ...
 <Directory /var/www/html/example.com>
    Option Indexes FollowSymlinks Multiviews
    AllowOverride None
    Require all granted
 </Directory>
  ...
```

Rajakavitha1:
Thanks for the issue @bagasme .
Just updated the guide in the Pull Request https://github.com/linode/docs/pull/2804.
After the team reviews the changes it will be published.
Closing this issue for now. Please feel free to reopen if any other updates need to be made.

PR #2804, opened 2019-10-18
[Update] Use NGINX as a Front-end Proxy and Software Load Balancer

Rajakavitha1:
- Updated the access control for apache 2.4 as perhttps://www.linode.com/docs/security/upgrading/updating-virtual-host-settings-from-apache-2-2-to-apache-2-4/
- Updated the IP address to use standard documentation IP address
- Fixes https://github.com/linode/docs/issues/2769


nmelehan:

Built with commit 132523f707478cc28bb08708d32014cee2f8b0ce

https://deploy-preview-2804--nostalgic-ptolemy-b01ab8.netlify.com

Rajakavitha1:
> Why example IP address also changed?

Hi @bagasme  we need to comply with https://tools.ietf.org/html/rfc3849 and https://tools.ietf.org/html/rfc5737 that is why the ip addresses are updated.

PR #2806, opened 2019-10-19
clarifications: locations of streisand.html on 2. Linode

viogq:
to clarify that a new, billed, Linode will be created. On this new Linode, streisand.html is to be found.


nmelehan:

Built with commit 9d3ad8bf0cdf49976c76955f2d9150ca81d0731d

https://deploy-preview-2806--nostalgic-ptolemy-b01ab8.netlify.com

viogq:
looks OK to me

Rajakavitha1:
Thanks for the PR @viogq !!!!!
We look forward to many such valuable contributions from you.

viogq:
the PR may be interesting for those who leapfrogged the first part and did not start from a local machine, but a Linode instead. Using a Linode is much safer since one does not have to allow to run a script with root. Much safer to do this on a new Linode, where it cannot harm the local system. And it works just as well. Besides, very few people run a local Ubuntu 16.04 in October of 2019.

PR #2863, opened 2019-11-06
[Update] How to Use Object Storage & Object Storage Pricing and Limit…

rsyracuse:
Added file size limit and info on multipart uploads

### DIFF
--- a/docs/platform/object-storage/how-to-use-object-storage/index.md
+++ b/docs/platform/object-storage/how-to-use-object-storage/index.md
@@ -160,7 +160,7 @@ You can drag and drop multiple files to the **Upload Files Pane** at one time.
     ![Successful upload of your object](successful-object-upload.png)
 
     {{< note >}}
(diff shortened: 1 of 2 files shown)


nmelehan:

Built with commit e2f5490680e3f4d3d72ca703e704e61fad79c3c6

https://deploy-preview-2863--nostalgic-ptolemy-b01ab8.netlify.com

hzoppetti:
If objects greater than 5GB can be uploaded via Multipart Uploads, is it accurate to say that individual or single files are limited to 5GB? Is this verbiage something that we were given? Is it better perhaps to say something like this:

Single file transfers are limited to 5GB each, however larger file transfers can be facilitated with Multipart Uploads. s3cmd and Cyberduck will do this automatically if an object exceeds this limit.

rsyracuse:
File transfers are only limited when uploading, and can be downloaded normally once they've been uploaded in multiple parts. Once uploaded in parts, they are rebuilt as the single object. It may be more accurate to say that "Single object uploads are limited" instead of files. Transfers in general aren't limited in this way, just uploads

PR #2876, opened 2019-11-11
Improvement/udf variable detail

lukelane124:
Added details about setting environment variable names inside a StackScript.

### DIFF
--- a/docs/platform/stackscripts/index.md
+++ b/docs/platform/stackscripts/index.md
@@ -270,13 +270,13 @@ Below is an example implementation of the UDF variables. Note that the UDF tags 
 {{< file "StackScript" bash >}}
 # [...]
 # <UDF name="var1" Label="A question" default="" example="Enter something here." />
-# ENVIRONMENT_VAR1=
+\# ENVIRONMENT_VAR1=
 # <UDF name="var2" Label="Pick one of" oneOf="foo,bar" example="Enter something here." />
-# ENVIRONMENT_VARNAME=
+\# ENVIRONMENT_VARNAME=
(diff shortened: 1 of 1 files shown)


nmelehan:

Built with commit d2370f8efd8e391eccd1fdeba2c410cbb2e3a167

https://deploy-preview-2876--nostalgic-ptolemy-b01ab8.netlify.com

Rajakavitha1:
@lukelane124  Thanks for your PR!!!

lukelane124:
@Rajakavitha1 No problem, glad to help!

nmelehan:
Hi @lukelane124 -

I don't think I fully understand the added lines from this PR--can you help me understand?

If you have a UDF variable defined, I can just use it immediately after. For example:

```
# <UDF name="var1" Label="A question" default="" example="Enter something here." />
echo $VAR1 > test.txt
```

I can see how this might not be clear in the guide, though, and I think we should add a line with an example like this.  Still, that's not quite the same thing as the lines you've added, so if you can help explain those to me I'd appreciate it.

I also am a little confused by some existing lines in the guide. Under the Creating a New StackScript section, we have:

```
#<UDF name="hostname" label="The hostname for the new Linode.">
# HOSTNAME=
#
#<UDF name="fqdn" label="The new Linode's Fully Qualified Domain Name">
# FQDN=
```

I think we could remove `HOSTNAME=` and `FQDN=`,  as the UDF line will make those variables available.

lukelane124:
@nmelehan In my limited experience using UDF in debian I was unable to access the UDF defined environment variables without the lines following each definition.
Also, sorry about the massive delay in reply time. I don't get on here much as work has been taking more than its fair share of my time.

If you need some clarification I'll see if I can pull something together.

PR #2971, opened 2019-12-13
Removing stale ref to UW resource

joemcl:
Removing invalid link to U Wisconsin resource - no longer on that page.

### DIFF
--- a/docs/tools-reference/linux-users-and-groups/index.md
+++ b/docs/tools-reference/linux-users-and-groups/index.md
@@ -12,7 +12,7 @@ modified_by:
 published: 2009-08-31
 title: Linux Users and Groups
 external_resources:
- - '[User Account and Group Management @ UWISC''s Center for Computer Aided Engineering](http://www.cae.wisc.edu/linaccounts)'
+ 
  - '[Users and Groups Administration in Linux @ DebianAdmin](http://www.debianadmin.com/users-and-groups-administration-in-linux.html)'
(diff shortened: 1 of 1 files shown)


docs/tools-reference/linux-users-and-groups/index.md
```diff
@@ -12,7 +12,7 @@ modified_by:
 published: 2009-08-31
 title: Linux Users and Groups
 external_resources:
- - '[User Account and Group Management @ UWISC''s Center for Computer Aided Engineering](http://www.cae.wisc.edu/linaccounts)'
```
bagasme: Broken link?

joemcl: @bagasme Yes but can't find the content on the page anywhere else, even

bagasme: After reviewing [web.archive.org snapshot](http://web.archive.org/web/20160926055501/http://www.cae.wisc.edu/cae-account-management/), actually it is account management home page for CAE course, not related to our Linux Users & Groups guide.

So removing this external resource is correct, to keep links related.

bagasme: Indeed, the external link above bounce to course account management page.

bagasme: Edit: It is actually broken link without redirection

joemcl: right

PR #3066, opened 2020-01-19
allow IPv6 /64 Pool addresses to work in CentOS7

hjpx:
Copying and pasting this configuration caused IPv6 /64 Pool addresses to silently fail in CentOS 7 for customers.

### DIFF
--- a/docs/networking/linux-static-ip-configuration/index.md
+++ b/docs/networking/linux-static-ip-configuration/index.md
@@ -121,7 +121,13 @@ PEERDNS=no
 # Edit from "yes" to "no".
 IPV6_AUTOCONF=no
 
-...
+# Keep the following lines:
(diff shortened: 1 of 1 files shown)


nmelehan:

Built with commit b68bc7617ece582e0c170eb267261ce162d511e0

https://deploy-preview-3066--nostalgic-ptolemy-b01ab8.netlify.com

hjpx:
Last minute suggestion:

The following seems like an invitation for the customer to use a pool address as their primary IPv6 address, which does not work.

```
# IPv6 gateway and primary address.
IPV6_DEFAULTGW=fe80::1%eth0
IPV6ADDR=2001:db8:2000:aff0::2/128
```

Perhaps we want to add the following to the comment line: "Primary address should be your SLAAC address."

PR #3084, opened 2020-01-21
Hotfix v.30.4

Guaris:
Navigation bar Change

Also including Leslie's spanish language shortcode from https://github.com/linode/hercules/pull/160


nmelehan:

Built with commit 25bf4eb47318b51a8f4939d161941618831ff7c4

https://deploy-preview-3084--nostalgic-ptolemy-b01ab8.netlify.com

nmelehan:
The call we make to `https://www.linode.com/wp-json/menus/v1/menus/submenu-products-developer-tools` is still returning `Users &#038; Permissions`, so we might want to hold off on merging this until that is fixed? This and the other navigation items are rendered when the site is built, which in our case is when the site is released to production.

If we wait til that is fixed, then merging and releasing this hotfix will also fix that issue.

If we don't wait to merge and release this, then the changes from this PR will go in, but the `Users &#038; Permissions` will still be present in the menu bar.

nmelehan:
Issues with the ampersand are resolved; the API no longer returns the encoding that we were seeing

PR #3110, opened 2020-01-31
[Hotfix] v0.31.1

nmelehan:
[Update] Linode Cloud Manager Security Controls
Removing alias that is the same as the guide's directory path

Reported by Linode support:
This guide is currently redirecting to itself in an infinite loop:
https://www.linode.com/docs/security/linode-manager-security-controls/

I noticed that one of the aliases on the guide is `security/linode-manager-security-controls/`, which is the same as that guide's path. I logged into the production server and saw that Hugo had rendered a redirect page at that URL, instead of a normal page with content.

Proposed fix is to remove that alias. Some other weird things I noticed:
- The last time we touched this line in the file was ten months ago, and we didn't even add this problematic alias at that time: https://github.com/linode/docs/pull/2414/files
- The problem isn't appearing for me when running on the dev server on my Mac (Hugo version 0.52), nor does it appear in the Netlify preview from #2414 
- I  am pretty sure that this guide has worked for the last ten months, so I'm not sure why it came up now.


nmelehan:

Built with commit 0e9e072b0cf4b0d07d4a5ba0c9fdb6e00874a10a

https://deploy-preview-3110--nostalgic-ptolemy-b01ab8.netlify.com

PR #3256, opened 2020-04-05
Update documentation on sticky bit

ErikSwan:
The current documentation for "Linux Users and Groups" seems to imply that the sticky bit, when set on _files_, will prevent users other than the owner or root from deleting the file, similar to the functionality when set on _directories_.

This is not correct—historically, the sticky bit, when set on files (specifically, executables), would save the `.text` section of an executable into swap space to improve performance for commonly-used executables. I'm not sure if this still technically happens with modern Linux distributions, but even if it does, the sticky bit has no practical use given the speed of modern storage devices.

The documentation should be corrected, as as it is currently written a user might think that setting the sticky bit on a _file_ that she owns within a shared directory would mean that only she could delete the file, but this is not the case (the sticky bit must be set on the enclosing _directory_, not on the file).

Sources:
- https://en.wikipedia.org/wiki/Sticky_bit
- The `chmod` man page:

      RESTRICTED DELETION FLAG OR STICKY BIT
             The restricted deletion flag or sticky bit is a single bit, whose interpretation depends on
             the file type.  For directories, it prevents unprivileged users from removing or renaming a
             file  in  the  directory  unless  they  own  the  file or the directory; this is called the
             restricted deletion flag for the directory, and is commonly found on world-writable  direc‐
             tories  like  /tmp.   For  regular files on some older systems, the bit saves the program's
             text image on the swap device so it will load more quickly when run;  this  is  called  the
             sticky bit.

### DIFF
--- a/docs/tools-reference/linux-users-and-groups/index.md
+++ b/docs/tools-reference/linux-users-and-groups/index.md
@@ -270,13 +270,15 @@ An octal table showing the numeric equivalent for permissions is provided below.
 
 In addition to the most common read/write/execute file permissions, there are some additional modes that you might find useful, specifically the *+t* mode (*sticky bit*) and the *+s* mode (*setuid bit*). These functions describe the behavior of files and executables in multi-user situations.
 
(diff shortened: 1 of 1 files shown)


nmelehan:

Built with commit f31eb2873c6706a84be00899b2e884d3b7b4f175

https://deploy-preview-3256--nostalgic-ptolemy-b01ab8.netlify.com

PR #3340, opened 2020-05-10
Minor Modifications and added some commands

gopalraha:
Replaced YUM with DNF as CentOS 8 is using DNF as a default package manager. Added "sudo ls -dZ /var/www/example.com" to check whether SELinux Context are added to the directory or not. Modified the command "sudo firewall-cmd --zone=public --permanent --add-service={http,https}" which is used to add the given services to firewall using single command this saves time. Added command "sudo firewall-cmd --list-all" to check the list of services and ports which are added to firewall.


nmelehan:

Built with commit be7673317ece63614e4b0f96f2e91488daa01a42

https://deploy-preview-3340--nostalgic-ptolemy-b01ab8.netlify.app

PR #3366, opened 2020-05-26
[Update] Control Network Traffic with iptables

Rajakavitha1:
updated based on a comment in DisQus:
"iptables-save > /etc/iptables/rules.v6 commands as any IPv6 changes will be overwritten by the IPv4 rules"

Sure it does if you use wrong command. The right one is ip6tables-save > /etc/iptables/rules.v6


nmelehan:

Built with commit 4270786bfe6addb575c2e6185a4271a08224b6fb

https://deploy-preview-3366--nostalgic-ptolemy-b01ab8.netlify.app

PR #3544, opened 2020-08-18
Clarifications for HTTP Cookie Session Stickiness

jerkob:
I was confused by the wording for the HTTP Cookie option and this is intended to clarify it.

It was unclear whether the HTTP Cookie method worked for the HTTPS protocol, and the description implied that subsequent routing would be based off of the client's IP rather than information in the cookie.


nmelehan:

Built with commit dd57d655148bf48bca44092c2b5c0dedc01fcebd

https://deploy-preview-3544--nostalgic-ptolemy-b01ab8.netlify.app

PR #3685, opened 2020-10-05
[Do not merge] URL rewrite test

bep:
/cc @nmelehan


nmelehan:
Actually, now that I look a little closer, the CSS and JS assets are not actually loading those things--clicking on them from the inspector brings up a badly-formatted HTML page too. going to keep investigating

bep:
Locally you need this for it to work: https://github.com/gohugoio/hugo/pull/7779 (that PR aligns "hugo server" with how Netlify works out of the box with redirects).

PR #3725, opened 2020-10-22
Fix comments on generate_permanent_redirects.sh

nmelehan:
The code was right, but the comments were "backwards"


wildmanonline (approved):

wildmanonline:
@nmelehan I don't believe we use this script any longer, but the change itself seems OK. Can you merge this in or close it?

PR #3726, opened 2020-10-23
[Update] Linode Beginner's Guide

leslitagordita:
* Move linode beginners guide to /guides/platform/get-started/ section
* @nmelehan want to confirm that we no longer need to alias guides when moving them, since all guides receive an auto-generated url `/docs/guides/guide-title`. I confirmed this locally, but I want to double-check.


nmelehan:
> @nmelehan want to confirm that we no longer need to alias guides when moving them, since all guides receive an auto-generated url /docs/guides/guide-title. I confirmed this locally, but I want to double-check.

@leslitagordita that's correct!

The only times we'll need to add an alias is:

- When we want to change the slug for the guide
- When we're moving a section to a different parent section, then the _index.md gets an alias from the old section path, because sections are not published under 'flattened' URLs

This will be part of the documentation I'm preparing ^

andystevensname (approved):

PR #4008, opened 2021-02-10
Update Algolia ranking configs

nmelehan:
These settings were accidentally not included in release v1.19.1.
They have been set correctly on the production Algolia index,
so this commit gets algolia_settings.json aligned with the
production settings.

### DIFF
--- a/algolia_settings.json
+++ b/algolia_settings.json
@@ -49,8 +49,8 @@
    "geo",
    "words",
    "filters",
-   "proximity",
    "asc(deprecated)",
+   "proximity",
    "attribute",
    "exact",
    "custom"
@@ -60,7 +60,7 @@
    "linode-documentation-sorted"
   ],
   "searchableAttributes": [
-   "title",
+   "unordered(title)",
    "keywords",
    "toc",
    "h1",


nmelehan:

Built with commit 3345bd39f31856d169e827f5f7f5adeeae4b0477

https://deploy-preview-4008--nostalgic-ptolemy-b01ab8.netlify.app

algolia_settings.json
```diff
@@ -60,7 +60,7 @@
    "linode-documentation-sorted"
   ],
   "searchableAttributes": [
-   "title",
+   "unordered(title)",
```
nmelehan, line 63: Setting the title search to unordered means that Algolia will not care whether the search term appears earlier in the title string:
https://www.algolia.com/doc/api-reference/api-parameters/searchableAttributes/

This matters because we haven't always had a strong convention for titles. For example:
- We may have a had a guide in the past that was titled "Apache on Debian"
- The new version of this guide might be titled "How to Install Apache Web Server on Debian"
- If the title search was ordered, then it would be possible for the earlier title to be prioritized in search rankings, because `Apache` appears at the start of that title.

algolia_settings.json
```diff
@@ -49,8 +49,8 @@
    "geo",
    "words",
    "filters",
-   "proximity",
    "asc(deprecated)",
+   "proximity",
```
nmelehan, line 53: This change makes it less likely that a deprecated guide will appear above a non-deprecated guide

andystevensname (approved):

PR #4024, opened 2021-02-15
[Hotfix] v1.21.1

leslitagordita:
Move to using the Linode docs theme repo

theme changes:
Always update search location when on topresults page
Crop the list thumbnails to fit their card container
Fix broken swipe on touch devices
Update module path

Update:
Fix h1 title for What is Unit Testing? guide.


bep (approved):

go.mod
```diff
@@ -8,4 +8,5 @@ require (
 	github.com/bep/hugo-jslibs/turbolinks v0.1.2 // indirect
 	github.com/bep/linodedocs v0.0.0-20210212231859-10aa00d2e096
```
nmelehan, line 9: @leslitagordita this PR looks good to me; the only thing I don't fully understand is why the old `github.com/bep/linodedocs` line remains in go.mod. But, that doesn't seem to be harming anything, and the theme files are correctly listed under `_vendor/github.com/linode/linode-docs-theme` instead of `_vendor/github.com/bep/linodedocs`. Also, I don't have a full understanding of what should be in go.mod and what shouldn't be.

@bep should we prune this line from go.mod, or does that not really make sense?

nmelehan (approved):

go.mod
leslitagordita: @nmelehan thanks for looking. I also had issues with this. I tried manually removing the line and it would come back. I tried `go mod edit -dropreplace` command to see if that removed the line permanently. Once I found the right config file to update with theme info, I saw the files where I expected them (and the old ones gone).

bep: Run `hugo mod tidy` and it should go away, but `hugo mod graph` tells the truth.

PR #4103, opened 2021-03-09
[Update] How to Install kubectl (shortguide)

alexfornuto:
The current download command assumes a download path that is no longer correct. I assume it has something to do with Google no longer being as active as it was with the K8s project.

You can validate the new path from the [Kubernetes docs](https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/).


andystevensname (approved):

alexfornuto:
image

PR #4543, opened 2021-07-23
Change configuration to enable "bundles" taxonomy

alertmybanjos:
I have tested this locally by checking out linode/linode-docs-theme/pull/89

This must be merged along with that pull request, or we'll get a build error that looks like this:
```
ERROR 2021/07/23 11:24:24 list.html for page "bundles" should not be used...
```


wildmanonline:
These changes were included in another PR.

PR #4700, opened 2021-09-07
Hide three guides from Explore Docs menu:

nmelehan:
These guides were appearing in Explore Docs menu, but don’t need to:

- /docs/guides/find-this-guide-on-github/
- /docs/guides/linode-writers-formatting-guide/
- /docs/guides/search/

guides in explore docs menu that should be hidden

Note that these menu items do not appear when loading the docs home page, but they do on other pages (like https://www.linode.com/docs/guides/). They also appear after a delay, after the other items in the menu show up.


nmelehan:
These will still show up in the Netlify preview for this PR, because the Netlify preview uses the production algolia index, and the Explore Docs menu is populated from Algolia. I have tested that the changes work by using a personal Algolia account.

bbiggerr (approved):

PR #4710, opened 2021-09-14
[update] added headers with images

Rajakavitha1:


docs/guides/applications/configuration-management/basics/an-introduction-to-infrastructure-as-a-service/index.md
```diff
@@ -9,6 +9,7 @@ keywords: ['infrastructure as a service','iaas','infrastructure','cloud networks
 tags: ['networking', 'linux']
 license: '[CC BY-ND 4.0](https://creativecommons.org/licenses/by-nd/4.0)'
 published: 2021-08-06
+image: An_introduction_to_ Infrastructure _as _a_ Service.jpg
```
nmelehan, line 12: @Rajakavitha1 The error from Netlify is showing that Hugo can't build successfully. I think it's indicating that there are some syntax issues with some of the new image frontmatter lines in this PR. 

The Netlify errors don't indicate which lines are the problem, but I skimmed through the PR and found this one, which has some extra spaces in the filename where it probably shouldn't. There might be some other problematic lines too, but I'm not sure. I'd recommend fixing this one and then trying to build with Hugo locally to see if it succeeds. If not, try looking through the other new image file names for similar issues

Rajakavitha1:
A new PR has been created to resolve the header issues https://github.com/linode/docs/pull/4738

PR #4910, opened 2021-11-22
Archive MariaDB tables to Object Storage

andysh-uk:


wildmanonline:
Title is very similar to an existing guide by the same author: https://www.linode.com/docs/guides/backup-mariadb-mysql-to-object-storage-with-restic/

andysh-uk:
Hi, the linked guide was about backing up a whole MySQL/MariaDB database to Linode Object Storage. The file saved on OS is a SQL file that can be used to restore the DB. The original table data still lives on the local server in MySQL/MariaDB on the original storage engine.

This guide in the PR uses Maria DB’s S3 storage engine to store the table on OS. Whenever the table is read from (e.g. SELECT * FROM …), the MariaDB storage engine reads the data from Object Storage. The data no longer lives on the server, it is stored in OS.

PR #4951, opened 2021-12-07
[Edit] Use Mypy for Python Static Type Checking

Sachin-Suresh:


wildmanonline:
This guide has already been released: https://www.linode.com/docs/guides/python-static-type-checking-with-mypy/

PR #5164, opened 2022-02-14
[Fix] Installing Apache Guacamole on Ubuntu and Debian

wildmanonline:
Fixed the command to copy the MySQL Connector/J (java) files from the extracted archive to the guacamole directory.


wildmanonline:
Closing this out in favor of #5145.

PR #5311, opened 2022-03-24
Removed duplicate aliases and the deprecated landing page guide

wildmanonline:
- Removed duplicate aliases
- Removed the deprecated landing page guide and either removed or updated the original guides.


wildmanonline:
Going to resubmit this to see if that fixes the no tests issue.

PR #5328, opened 2022-03-31
[New] How Many Kilobytes are in a Megabyte?

leslitagordita:
Preview: [Data Storage Guide: How Many KB are in a MB?](https://deploy-preview-5328--nostalgic-ptolemy-b01ab8.netlify.app/docs/guides/how-many-kb-in-a-mb/)

### DIFF
--- a/docs/guides/quick-answers/linux/how-many-kb-in-a-mb/index.md
+++ b/docs/guides/quick-answers/linux/how-many-kb-in-a-mb/index.md
@@ -1,0 +1,81 @@
+---
+slug: how-many-kb-in-a-mb
+author:
+  name: Jack Wallen
+description: 'How many KB are in a MB? Which is bigger, MB or GB? Data units can present lots of questions. Read our guide to become a data storage expert.'
+keywords: ['how many kb in mb','mb','kb to gb','megabytes in a gigabyte']
+license: '[CC BY-ND 4.0](https://creativecommons.org/licenses/by-nd/4.0)'
+published: 2022-03-31
+modified_by:
+  name: Linode
+title: "How Many Kilobytes are in a Megabyte?"
+h1_title: "Data Storage Guide: How Many KB are in a MB?"
+enable_h1: true
+contributor:
+  name: Jack Wallen
+---
+
+Kilobytes and Megabytes are units of measure that are used in nearly every computing platform on the planet. This goes for mobile phones, desktops, servers, virtual machines, containers, and cloud hosting. Being well-versed in these measurements is important in a variety of applications, but is especially important when you work with third-party cloud vendors that charge for bandwidth and storage . At a bare minimum, you need to calculate the average usage so you know how much that cloud service is going to cost.
+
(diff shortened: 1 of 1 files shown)


docs/guides/quick-answers/linux/how-many-kb-in-a-mb/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+In the early days of the Internet, you had a modem that was either 14.4, 28.8, or 33.6. Those numbers represented how many bits per second the hardware could send or receive. So 14.4, 28.8, and 33.6 were actually measured in kilobits per second (or kbits/s), so 14.4 kbits/s, 28.8 kbits/s, and 33.6 kbits/s.
+
+A byte is considered a group of 8 bits, so 1 byte is equal to 8 bits. Historically, a byte was the number of bits used to encode a single character of text. Think about it this way: The binary code for the letter "a" is 01000001 and 01000001 is 8 bits. So each character is represented by 8 bits.
+
+One thing to keep in mind is bits are represented by a lower case "b" and a byte is represented by an upper case "B."
+
+## What Are Kilobytes?
+
+A kilobyte is approximately 1,000 bytes and since a byte is 8 bits, a kilobyte is roughly 8,000 bits. But why do we say "approximately" and "roughly"? A kilobyte is actually 1024 bytes.
```
wildmanonline, line 43: A kilobyte is 1000 bytes. A kibibyte is 1024 bytes. https://www.google.com/search?q=how+many+bytes+in+a+kilobyte&rlz=1C5CHFA_enUS998US998&oq=how+many+&aqs=chrome.0.69i59l3j69i57j69i59j69i61l2j69i60.1488j0j7&sourceid=chrome&ie=UTF-8

wildmanonline:
I think the content of this article should be revisited. This is a fairly confusing topic (storage units are a mess!) and it would do well to have clear, accurate, and concise answers within our documentation. A common reason people seek more information on these units of storage is to understand the difference between base 10 (KB, MB, GB, etc) and base 2 (KiB, MiB, GiB, etc). This is especially confusing because operating systems and services often use base 10 unit designations to describe base 2 data. 

Is a kilobyte 1000 bytes or is it 1024 bytes? There's varying answers out there and, in my opinion, it's beneficial to cover the common consensus as well as the actual standards of measurement. KB, GB, TB are base 10 as defined by standards. They sometimes are used to describe base 2, though KiB, GiB, and TiB were created as base 2 storage units to make a clear distinction. Adoption and public knowledge of KiB, GiB, and TiB isn't widespread and, as such, leads to the general public being confused about storage units.

As an example, within File Explorer, Windows calculates sizes using base 2 (GiB) but displays units as base 10 (GB). This means a 500 GB drive might appear as 465 GB in Windows. The size of the drive hasn't changed, it's just that Windows isn't using the standard (which would be 465 GiB or 500 GB - *not 465 GB*).

I haven't researched this topic for a few years now, so I could have misspoke on a point or two, though this is my current understanding of storage units.

PR #5386, opened 2022-04-25
[Update] Managed Databases > Automatic Updates and Maintenance Windows

wildmanonline:
Reverted change related to DBMS version upgrades. This now clarifies that *minor version releases* are not included in the automatic updates that occur during maintenance windows.


bbiggerr (changes requested):
The exclusion of major versions here makes this confusing -- there's an implication that major version updates are included, but not minor. Would be clearer to state that there are no engine updates included of any kind. All engine updates must be accomplished through manual migration at this time.

nmelehan (approved):

PR #5467, opened 2022-05-28
[Update] Getting Started with the Linode API

Wat-Poods:
We no longer provide xen hypervisors (haven't since before when I started with Linode in the summer of 2018. Edited the example API calls for what I _think_ it should be, though haven't tested the command.


wildmanonline:
Thanks @Wat-Poods ... this guide needs a complete overhaul as much of that content is outdated! The code block you were referencing was actually just the output of a public API call so I reran that call and copied in the current output. I'll create a task for us to rewrite this guide soon.

wildmanonline (approved):

PR #5490, opened 2022-06-06
Use scroll-padding-top to offset sticky header when clicking anchor links

bavedarnow:
Adding the scroll-padding-top property will accommodate for the height of the sticky header when clicking anchor links. I haven't tested this code, but adding the property to the html element in chrome devtools produces a desirable result

CleanShot 2022-06-06 at 11 58 17

The calculation would need to pull the `--height-navbar-row` and add an additional padding, like a line-height variable, if that exists. See this guide for more options: https://getpublii.com/blog/one-line-css-solution-to-prevent-anchor-links-from-scrolling-behind-a-sticky-header.html


wildmanonline:
@bavedarnow This issue has been imported over to our theme repo. Closing out this PR. Thanks again!

PR #5560, opened 2022-07-05
Vale improvements (including temporary fix for possessive dictionary items)

wildmanonline:
This PR includes the following changes to Vale:

- Set Vale to only report errors to make Github Action log easier to read and pinpoint errors. This doesn't affect our Visual Studio Code reporting, but does impact if we run Vale manually (which I think is acceptable behavior).
- Updated Github Action to use the newest action version and use Vale CLI version v2.19.0
- Updated our custom spelling rule to filter out (ignore) any possessive words (reverting a change that was made to Vale in v2.18.0). Specifically, we're adding back in the filter `\W`.
- Added a Repetition rule to identify words that are repeated (such as "the the")


wildmanonline:
This PR can be merged if we want to continue ignoring possessive dictionary terms. Here's my thoughts on the issue:

- Update the spelling rule to ignore possessives of dictionary terms so we can continue as usual without interruption. Then think up a longer term solution later.
- Leave the spelling rule as is. When Vale alerts us to a possessive word failure, add it to the dictionary/ignore list.

The only other minor consideration with this PR is related to the Repetition rule. In our entire library, there is just one case of a false positive (where the correct term is a repeated word): "Tiny Tiny". My thought is that we can add exceptions to this within the Repetition rule as they come up, just like we add exceptions in our dictionary.txt file for spelling. If this is what we want to do, I can build that out in this PR or another PR shortly after.

andystevensname (approved):

PR #5573, opened 2022-07-11
[Update] Running a Mail Server

jschauma:
The current text makes it seem as if using a self-signed certificate was a reasonable choice. With the proliferation of free, easy-to-renew certificates such as Let's Encrypt, it really seems that we can more strongly encourage good security practice instead of training people to deploy self-signed certs and click through certificate warnings, so this PR nudges users a bit more strongly to use a trusted CA.


jfrederickson:
The guide linked to here is specifically about purchasing and using a commercial TLS cert rather than using an ACME-provided cert.

Personally, I'd rather encourage people to use one of the free providers these days as you mention above, but it looks like the [Let's Encrypt doc](https://www.linode.com/docs/guides/install-lets-encrypt-to-create-ssl-certificates/) we have is considered deprecated in favor of a web-specific doc that doesn't quite apply here. Linking to something like that would be my preference, but linking to a deprecated doc would feel weird...

jschauma:
Yeah, I had specifically changed the wording from "purchase a signed certificate" to "request a trusted certificate" to avoid favoring a commercial solution.  Changing the content of the linked doc at https://www.linode.com/docs/guides/obtain-a-commercially-signed-tls-certificate/ (which does include LE explicitly) is something that might make sense but it outside the intended scope of this PR.

wildmanonline (approved):

wildmanonline:
Thanks @jschauma. Tweaked your changes a bit and then added back in information on responsibly using self-signed certificates (internal-only usage really). This should go live later today.

jschauma:
Great, thanks!

PR #5624, opened 2022-08-04
Fix a dead link to cert-manager helm chart page

jahrlin:
We recently moved cert-manager from the sponsoring org.

This PR fixes a dead link that was a result from that


wildmanonline (approved):

wildmanonline:
Thanks for the PR @jahrlin. I'll merge this in now and it should be live later today.

PR #5639, opened 2022-08-09
Updates to Managed Database guides (note about resize)

wildmanonline:
This PR adds a note to several guides to notify users that Managed Databases can't be resized once created.

- [Update] Managed Databases > Create a Managed Database
- [Update] Managed Databases > Database Engines and Plans

### DIFF
--- a/docs/products/databases/managed-databases/guides/create-database/index.md
+++ b/docs/products/databases/managed-databases/guides/create-database/index.md
@@ -4,7 +4,8 @@ author:
   email: docs@linode.com
 title: "Create a Managed Database"
 description: "A walkthough on creating a Managed Database through the Cloud Manager"
-modified: 2022-06-06
+published: 2022-06-06
+modified: 2022-08-09
 ---
 
 This guide walks you through creating a Linode Managed Database through the Cloud Manager.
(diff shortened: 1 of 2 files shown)


docs/products/databases/managed-databases/guides/create-database/index.md
```diff
@@ -48,6 +49,10 @@ Every node of a database cluster is built on its own Linode Compute Instance. In
 
 - [Choosing a Compute Instance Type and Plan](/docs/guides/choosing-a-compute-instance-plan/)
 
+{{< note >}}
+Once a Managed Database cluster is created, it cannot be resized to a different plan. To modify the resources allocated to your database, you will need to create a new Managed Database with the desired plan, migrate your data, and delete the original Managed Database.
```
nmelehan, line 53: am I overthinking this, but maybe it would be beneficial to link to the migrate database guide here
https://www.linode.com/docs/products/databases/managed-databases/guides/migrate-mysql/

And a link to a deleting a database guide if we have that (not finding one right now)

nmelehan (approved):

docs/products/databases/managed-databases/guides/create-database/index.md
wildmanonline: Thinking about this now, we don't have a single migration guide. We have one for each database engine, so might be too verbose to link to each one. Since this lives directly alongside those guides (so they should already be visible in the explore docs menu), going to leave out the links for now.

PR #5751, opened 2022-09-23
[UPDATE]-5726-5727-5734-5741

adamovera:
This is an update to PR #5726 #5727 #5734 and #5741 to replace a number of code boxes with output boxes.


wildmanonline:
Just updated some of those outputs to file shortcodes. Good to go!

wildmanonline (approved):

PR #5767, opened 2022-09-29
[Update] Configure an Email Server with Postfix, Dovecot, and MySQL on Debian and Ubuntu

wildmanonline:
This PR notifies the reader that the SHA2() function in MySQL 5.8 and above should not be used when generating passwords. It also cleans up the `doveadmin pw` command and fixes some formatting issues.


jahschwa (approved):
I haven't tested this, but the changes to wording are great security-wise :+1:

andystevensname (approved):

PR #5800, opened 2022-10-08
Fix link to updated guide

Wat-Poods:
The link to the "Manual Network Configuration on a Compute Instance" guide doesn't work. It looks like the link is missing the /docs/ directory in the URL.


Wat-Poods:
Oh, this is actually the same thing as PR 5799, @eliasvc beat me to this https://github.com/linode/docs/pull/5799

PR #5916, opened 2022-11-29
[Update] An Overview of IPv6 on Linode

nullfinity:
The previous edition of this document suggested but did not make explicitly clear that using a /64 or /54 routed range requires disabling Network Helper on the Linode and manually configuring its network settings for IPv6 addresses from its pool range. This PR addresses this potential pitfall by making this need explicitly clear.


wildmanonline (approved):

wildmanonline:
Thanks for the PR @nullfinity! Made a few minor copy changes but this looks good to me.

PR #5924, opened 2022-12-01
New test that checks markdown links for issues

wildmanonline:
This PR introduces a test that checks all markdown links within our guides and determines if there are any issues. It also checks for duplicate aliases.

Here are the links that are ignored:
- Correctly formatted links to external sites

Here are the types of issues that are flagged by this test:
- **Target not found** (FAILURE): *A link does not point to a guide or alias of a guide*
- **Duplicate aliases** (FAILURE): *The same alias has been added to multiple guides*
- **Slug mismatch** (FAILURE): *The guide's slug does not match the name of its parent folder*
- **Contains the linode docs domain** (FAILURE): *A link contains `linode.com/docs`*
- **Incorrect root** (FAILURE): *A link points to a guide but does not start with `/docs`*
- **Formatting issues** (FAILURE): *A link contains multiple slashes or does not start with a slash.*
- **Points to alias** (WARNING): *A link points to an alias of a guide. This is the only issue that does not result in a failure. This means we should run the update-internal-links.py script.* `python3 ./scripts/update-internal-links.py`

Command to run script locally (when in the top level directory of your local repo):

```
python3 ./ci/check-links.py
```

Optionally add the `--fix_issues` argument to attempt to fix *some* issues. Currently the only issues it can fix are slug mismatches.


wildmanonline:
View the test results for the latest commit. There are several accurate failures shown that already exist in our library.

nmelehan (approved):

PR #5984, opened 2023-01-04
Remove h1_title in favor of title_meta

wildmanonline:
This PR contains the following changes:

- If a guide uses the `h1_title` parameter, the `title` parameter becomes `title_meta` and the `h1_title` parameter becomes `title`.
- Removes the `h1_title` and `enable_h1` parameters from all guides as well as from the `yaml_rules.json` file. If a PR contains either of these parameters, the associated test should fail.
- If `title` and `title_meta` contain the same text, `title_meta` is removed from the guide.
- If `description` and `og_description` contain the same text, `og_description` is removed from the guide.

The script that performed most of the above actions is stored within `scripts/misc/update-frontmatter.py`. This is mainly for archive purposes and so that we can reuse the code if we need to perform other similar actions in the future.


wildmanonline:
Closing in favor of #5987 (updated script to fix some unintended behavior)

PR #6053, opened 2023-01-31
[update] Create an Email Server using Mail-in-a-Box

Rajakavitha1:
Updated the version of Ubuntu based on a comment in DisQus:
```
Taken from the Mail-in-a-Box homepage: https://mailinabox.email/

<i>Note: The video is a bit old. You now must select a server running Ubuntu 22.04 x64. The other steps remain almost entirely the same, so you can still use this video for reference.</i>

So you may want to change bullet 3 under <b>Before You Begin</b>: https://www.linode.com/docs/guides/mail-in-a-box-email-server/#before-you-begin
```


wildmanonline:
@Rajakavitha1 has this been tested yet on Ubuntu 22.04 LTS? If yes, we can merge it in. If not, do we want to close it out or keep it open for now?

Rajakavitha1:
> @Rajakavitha1 has this been tested yet on Ubuntu 22.04 LTS? If yes, we can merge it in. If not, do we want to close it out or keep it open for now?

Hi @wildmanonline It has been a while and I don't seem to have the terminal output that I usually save after validation for this PR. May be we can merge this because the user who reported this does mention that it works for the updated version of Ubuntu.

wildmanonline (approved):

PR #6113, opened 2023-02-28
feat: update appwrite images

christyjacob4:
Updated the Getting Started with Appwrite Tutorial with newer screenshots & updated Docker image version


wildmanonline:
Thanks for this PR! Before we can merge it, we do need updated screenshots that do not show a full name on the top left (currently "Walter O'Brien"). This can be replaced with either "Example User" by editing the source or it can be blurred out. We can also take care of this at some point down the road, but welcome any additional commits you can make.

wildmanonline (approved):

PR #6349, opened 2023-05-16
Add region ID to the availability table

zliang-akamai:
This would make the region list nicer for the developers who want to directly apply the region ID to their code.

Do you guys think this is a good idea?


zliang-akamai:
Feel free to close this PR if it wasn't a good suggestion.

PR #6393, opened 2023-06-04
Update index.md

ryaeng:
`terraform output kubeconfig` includes double quotes by default. Adding ` | cut -d \" -f 2` pipes the output through to cut and removes the double quotes.


ryaeng:
This is incorrect. Typo. Need to change it.

PR #6500, opened 2023-07-25
[Update] How to Migrate a CPanel Server to Linode

Rajakavitha1:
- Added the option to deploy cpanel using marketplace
- Updated the other distros that cn be used.


wildmanonline:
I reworked the following pieces of content:

- **How to Migrate a cPanel Server to Linode > Deploy cPanel on Linode** - I restructured this to better outline both methods of deploying cPanel on Linode (Marketplace App and manual installation).
- **Install cPanel on CentOS > Before You Begin** - Added this section in as there was no place in this guide that discussed actually deploying an instance. This content was previously in the migration guide, but didn't make sense there with the changes above.

There's some other miscellaneous cPanel related changes as well.

PR #6509, opened 2023-07-28
New CI/CD GitHub Actions workflows

nmelehan:
New workflows in `.github/workflows/`:
- deploy-to-production
- deploy-to-main-staging-server
- deploy-to-feature-testing-server
- update-algolia-on-schedule


.github/workflows/deploy-to-production.yaml
```diff
@@ -0,0 +1,218 @@
...
+
+        rsync_to_destination "${{ secrets.DOCS_WEBSITE_RSYNC_DESTINATION_1 }}"
+        rsync_to_destination "${{ secrets.DOCS_WEBSITE_RSYNC_DESTINATION_2 }}"
+
+        rm rsync_known_hosts
+
+    - name: Update Algolia
+      working-directory: ./linode-docs-theme-repo
+      env:
+        ALGOLIA_APP_ID: ${{ vars.ALGOLIA_APP_ID }}
```
nmelehan, line 199: Note that I have created a separate testing Algolia app (also separate from the sandbox) for this production-update workflow so that we don't interfere with the actual production Algolia app for now.

nmelehan: This separate Algolia app is now failing to be updated in the production workflow, because it hit its usage limits for the free tier it's on. But, it was working for a few runs, so this isn't concerning to me right now

wildmanonline (approved):

PR #6523, opened 2023-08-07
[update] Rescue and Rebuild

Rajakavitha1:
_ clarify that the additional disks are not mounted by default.


jddocs (approved):

jddocs:
Changes made based on feedback:

- Note about disks needing to be mounted manually during the "Add Disk" section w/ a link pointing to Mounting Disks section lower in the page
- Took most of the Step 5 note and moved them into the regular instructions for that step
- Clarified naming disk conventions as a "best practice"

PR #6536, opened 2023-08-14
90 day OTP details added to 2FA and Security Control guides

jddocs:
Added notes to the 2FA and Security Control guides re: OTP codes sent to users who haven't logged in in 90 days.

Also updated note formatting for managing 2FA through third parties.


wildmanonline (approved):

wildmanonline:
Made some changes to the note on the 2FA page:
- Changed placement --- moved it to the top of the page so that users understand what protections are in place should they not enable 2FA.
- Added a title: "Account protection for users without 2FA enabled"
- Changed some of the wording to clarify a few things. 
- Removed sentence that mentioned enabling 2FA bypasses this feature. Bypassing security features almost makes is sound as if 2FA is less secure. It's more of an enhancement, so I think we can do without that.

PR #6601, opened 2023-09-11
Added disk UUID notes to Config Profile and Backup Restore guides

jddocs:
Added:
- Configuration profiles guide: Note in Block Device Assignmetn about UUID disk assignments in fstab
- Restore to an Existing Instance guide: Warning about mounting backup root disks to the original configuration profile


docs/products/storage/backups/guides/restore-to-an-existing-instance/index.md
```diff
@@ -33,14 +33,22 @@ This process restores all data that was stored on the disk at the time the backu
...
+
+    {{< note type="warning" title="Warning: UUID Collisions">}}
+    When you restore a backup, the restored disk is assigned the same [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier) as the original disk. In most cases, this is acceptable and does not cause issues. However, if you attempt to mount both the original disk and the corresponding restore disk at the same time (by assigning them both to devices in your Configuration Profile's **Block Device Assignment**), you will encounter a UUID "collision".
+
+    When this happens, the system selects, and mounts, only one of the disks at random. This is due to both disks sharing the same UUID, and your instance *may fail to boot* since it will not be clear which disk is root. If your system does boot, you will not see any immediate indication if you are booted into the restored disk or the original disk, and you will be unable to access both disks at the same time.
```
wildmanonline, line 49: Approving this PR, with the condition that the following is true:

> This is due to both disks sharing the same UUID, and your instance *may fail to boot* since it will not be clear which disk is root. If your system does boot,...

If it's not true, tag me in again once there's a copy edit.

wildmanonline (approved):

PR #6730, opened 2023-11-14
[Update] Transfer Services to a Different Account

Rajakavitha1:
_ added a note about ensuring that the compute instance is not a part of the VLAN

### DIFF
--- a/docs/products/platform/accounts/guides/service-transfers/index.md
+++ b/docs/products/platform/accounts/guides/service-transfers/index.md
@@ -29,6 +29,7 @@ Only users with full account access can use Service Transfers. To adjust user pe
 
 {{< note type="warning">}}
 Each Compute Instance that you wish to transfer must have a unique label that is not used for any other instances on the receiving account. If the label is not unique, the transfer will fail. As an example, this commonly happens when the default label has been used. To ensure that the transfer is successful, verify that the label of the Compute Instance you are transferring is unique on the receiving account.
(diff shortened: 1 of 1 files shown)


docs/products/platform/accounts/guides/service-transfers/index.md
```diff
@@ -29,6 +29,7 @@ Only users with full account access can use Service Transfers. To adjust user pe
 
 {{< note type="warning">}}
 Each Compute Instance that you wish to transfer must have a unique label that is not used for any other instances on the receiving account. If the label is not unique, the transfer will fail. As an example, this commonly happens when the default label has been used. To ensure that the transfer is successful, verify that the label of the Compute Instance you are transferring is unique on the receiving account.
+Also, ensure that the Compute Instance is not configured to be in VLAN.
```
wildmanonline, line 32: If this is a hard blocker, it should be added to the corresponding lists further down the page instead of this section.

> -   Compute Instances must not:
>     - be assigned to a NodeBalancer, Firewall, VLAN, or Managed Service.

If the transfer is still allowed, but there are negative customer-facing impacts, those impacts should be presented to the reader here. In addition, we should avoid single sentence paragraphs when possible.

Rajakavitha1: Thank you @wildmanonline . Incorporated the changes that you suggested. I added it before the actual steps to transfer the services because I thought it was a prereq. The section that you suggested is appropriate.

wildmanonline (approved):

PR #6731, opened 2023-11-15
[Update] Use Glances for System Monitoring

timzl:
Changed two instances of '61208' to '61209' in the 'Setting up Clients and Servers' section so to accurately reflect the port that is open by default


Rajakavitha1 (approved):
LGTM!

wildmanonline (approved):

PR #6735, opened 2023-11-18
[update] Guides - Setting Up and Securing a Compute Instance

Rajakavitha1:
Added a note about the additional command that need sto be run to start the sshd service.


wildmanonline (changes requested):
On Ubuntu 22.04, running the regular `systemctl restart sshd` command seems to work (at least when disabling root ssh access). When does this command not work and when is the new `systemctl enable --now ssh.service` needed?

If the command is needed for certain distributions, we should pull it out of the note and put it in the instructions.

wildmanonline (approved):

PR #6954, opened 2024-05-15
Use `mariadb` client instead of `mysql` in `docs/guides/how-to-install-mariadb-on-debian-10`

h3ssan:
## Reason for the change

- To align with the usage of `mariadb` client as it used by the official MariaDB website.
- This change helps remove confusion by using consistent terminology throughout the documentation.

## What Official MariaDB says?

Please [`check this article`](https://mariadb.com/docs/server/connect/clients/mariadb-client/#Client_10.3_and_Older) as it said:
```
Modern versions of MariaDB Client use client binary filename of mariadb

For MariaDB Client 10.3 and older, the client binary filename was mysql rather than mariadb.
```

Note that the LTS version of MariaDB is `10.11`!

## Testing

- I checked the page manually and confirmed it looks great

### DIFF
--- a/docs/guides/databases/mariadb/how-to-install-mariadb-on-debian-10/index.md
+++ b/docs/guides/databases/mariadb/how-to-install-mariadb-on-debian-10/index.md
@@ -55,13 +55,13 @@ Allowing unrestricted access to MariaDB on a public IP not advised but you may c
 
 ### MariaDB Client
 
-The standard tool for interacting with MariaDB is the `mariadb` client, which installs with the `mariadb-server` package. The MariaDB client is used through a terminal using the `mysql` command.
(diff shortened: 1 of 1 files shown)


wildmanonline (approved):

wildmanonline:
Thanks for the contribution @h3ssan! We'll get this into the next release.

h3ssan:
> Thanks for the contribution @h3ssan! We'll get this into the next release.

Hi @wildmanonline 
I see that the latest release is out 2 days ago, but my pr still opened.

wildmanonline:
@h3ssan That was a one-off release outside of our normal release cadence. Rest assured, it will be release soon, likely at the next normal release (scheduled for tomorrow but that could change).

PR #6963, opened 2024-05-20
Fix bad anchor links

wildmanonline:
This PR addresses bad anchor links, specifically those that do not have a corresponding heading/anchor on the destination page.


wildmanonline:
Vale test can be ignored. It's failing due to too many files.

jddocs (approved):

PR #7013, opened 2024-06-21
Fix mistyped "manager" in various documentation.

jcolp:
This change does a pass through the documentation for misuse of "manger" instead of "manager".


wildmanonline (approved):

wildmanonline:
Thanks for the PR @jcolp!

PR #7033, opened 2024-07-05
Fix outdated information on the NixOS guide

### LINKED ISSUE #7032: Install and Configure NixOS on a Linode Proposed Changes
Link: https://www.linode.com/docs/guides/install-nixos-on-linode/
## Issue
In this guide, we create an install disk for NixOS with a size of ```1024 MB```. For older versions of NixOS, this worked fine, but with the latest version of NixOS (24.05), the ISO has grown past ```1024 MB```. This is causing an issue with curl not being able to fetch the entire ISO due to it exceeding disk space.

## Suggested Fix
Increase the installer disk size from ```1024 MB``` to ```1280 MB```. This give some slack for future updates to the ISO as well.
The installer disk is also recommended to be deleted at the end of the guide, so initial size should not matter much as long as it is big enough to copy NixOS.

21CSM:
This change modifies the guide to increase the default installer disk size from 1024 MB to 1280 MB, due to latest NixOS minimal ISO size being over 1024 MB.

Fixes #7032.

### DIFF
--- a/docs/guides/tools-reference/custom-kernels-distros/install-nixos-on-linode/index.md
+++ b/docs/guides/tools-reference/custom-kernels-distros/install-nixos-on-linode/index.md
@@ -5,7 +5,7 @@ description: 'Install NixOS, which is known for its declarative approach to conf
 authors: ["Andrew Miller"]
 contributors: ["Andrew Miller"]
 published: 2017-06-16
-modified: 2023-07-11
+modified: 2024-07-05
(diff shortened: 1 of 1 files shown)


Rajakavitha1 (approved):

21CSM:
Would it be possible to merge this @nmelehan?

jddocs (approved):

PR #7100, opened 2024-09-13
Remove Content-Security-Policy from the Hugo server config

bep:
Note that this is the `hugo server` only; it's getting out of date (re. consent manager), so instead of spending time maintaining it, just remove it.


bep:
@nmelehan You need to merge this.

PR #7110, opened 2024-09-27
Fix some invalid front matter dates

bep:
These will fail the build in the next Hugo version, but should be fixed even if you never plan to upgrade.


nmelehan:
@bep - thank you

nmelehan (approved):

Issue #3153, opened 2020-02-25
How to Access PostgreSQL Database Remotely Using pgAdmin on Windows


Rajakavitha1:
Update the guide based on the comment in DisQus:
Please update to pgadmin 4 and add SSH tunneling method thru pgadmin

PR #7150, opened 2024-11-21
[update] Securely Manage Remote PostgreSQL Servers with pgAdmin on Mac

Rajakavitha1:
updated the guide to include instructions using the inbuilt SSH tunneling in pgAdmin 4

Fixes: https://github.com/linode/docs/issues/3153


docs/guides/databases/postgresql/securely-manage-remote-postgresql-servers-with-pgadmin-on-macos-x/index.md
```diff
@@ -18,62 +18,76 @@ tags: ["database","postgresql"]
...
+   - **Maintenance Database**: `postgres` or your database name.
+   - **Username**: Your PostgreSQL username. For example: `postgres`
+   - **Password**: The password for your PostgreSQL user.
+5. Navigate to the **SSH Tunnel** tab:
+   - **Use SSH tunneling**: Enable this option.
+   - **Tunnel host**: Your Linode's IP address.
+   - **Tinnel port**: `22` . This is the default SSH port.
+   - **Username**: Your SSH username for the Linode instance.
+   - **Authentication**: Choose `Identity file` if you are using an SSH key, or `Password` for password-based authentication.
+   - **Identity file**: If you are using an SSH key, browse and select the private key file.
```
jddocs, line 65: @Rajakavitha1 is this accurate? I asks to select a "private key". Is this actually referring to a user's pubic key they would need to provide?

Rajakavitha1: Hi @jddocs  I checked the official documentation and this is what it reads: https://www.pgadmin.org/docs/pgadmin4/development/server_dialog.html
```Specify the type of authentication that will be used when connecting to the SSH host in the Authentication field:

Select the Password option to specify that pgAdmin will use a password for authentication to the SSH host. This is the default.

Select the Identity file to specify that pgAdmin will use a private key file when connecting.```

The users public key is uploaded when the hardening the [SSH access](https://techdocs.akamai.com/cloud-computing/docs/set-up-and-secure-a-compute-instance#harden-ssh-access) even before PostgreSQL is installed on Linode.  My understanding is that it is the private key that  is used to securely establish an SSH connection to the server. The server verifies this key against the public key stored in the ~/.ssh/authorized_keys file on the Linode instance.

jddocs (approved):

PR #7203, opened 2025-02-27
[NEW] Comparing CPUs, GPUs, and VPUs guide

jddocs:
New guide reviewing hardware differences between CPUs, GPUs, and VPUs


jddocs:
doc is no longer going to be published, converting content to blog instead. closing PR.

PR #7233, opened 2025-04-08
Update required python version

bluesmoon:
The requirements.txt file includes Ansible 10.4 which only works with Python 3.11 or higher, so this PR updates the required python version to 3.11.


nmelehan-akamai:
@bluesmoon thank you!

nmelehan-akamai (approved):

PR #7318, opened 2025-08-08
Update template to match changes to the Linode Writer's Formatting Guide

DHBR2:
Just one change in the link from http to https.


DHBR2:
Discarded.

PR #7429, opened 2026-03-30
Retire-Install-Apache-on-CentOS8

DHBR2:
Retire EOL OS; In the future replace with a new guide for installing Apache on Rocky Linux 9 or Alma Linux 9 and there is another guide "Use Certbot to Enable HTTPS with Apache on CentOS 8" (published July 2021, https://www.linode.com/docs/guides/enabling-https-using-certbot-with-apache-on-centos-8/ ) - which mentions Rocky Linux 8 and AlmaLinux 8 in its description, indicating awareness of the CentOS transition. That one should be updated to 9 also.


DHBR2:
Duplication.

